PR #22, opened 2021-03-30
Consume microsoft vscode-extension-telemetry

ameyaapte1:
Consime vscode-extension-telemetry package to log event telemetry from the vscode plugin


davidjenni:
obsoleted by #48

PR #48, opened 2021-05-25
Integrate with App Insights telemetry

joem-msft:
Integrate with `vscode-extension-telemetry` and App Insights to start tracking usage and the major lifetime events.


src/client/lib/CliAcquisition.ts
```diff
@@ -61,6 +63,7 @@ export class CliAcquisition implements IDisposable {
     async installCli(pathToNupkg: string): Promise<string> {
         const pacToolsPath = path.join(this._cliPath, 'tools');
         if (this.isCliExpectedVersion()) {
+            this._context.telemetry.sendTelemetryEvent('PacCliAcquired', { cliVersion: this.cliVersion, installState: 'AlreadyInstalled' });
```
davidjenni, line 66: that is likely too noisy: this code path is walked on every extension activation, e.g. when opening or reloading a VSCode window.  -> remove

src/client/telemetry/localfileusersettings.ts
```diff
@@ -0,0 +1,53 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+
```
davidjenni, line 2: nit: missing MIT license comment

src/client/telemetry/localfileusersettings.ts
```diff
@@ -0,0 +1,53 @@
...
+    const userSettingFilePath = path.join(applicationPath, userSettingsFileName);
+    if (fs.existsSync(userSettingFilePath)) {
+        const stream = fs.readFileSync(userSettingFilePath, 'utf8');
+        try {
+            settings = JSON.parse(stream) as IUserSettings;
+            if (settings.uniqueId === '') {
+                writeUserSettings(settings, applicationPath);
+            }
+        } catch (err) {
+            fs.unlinkSync(userSettingFilePath);
```
davidjenni, line 36: nervous that this will throw if the file cannot be deleted... thus disabling this vscode extension for the full duration of the session.
Better to fail graceful by either using the async variant (which would turn this func into async, sigh...) or for now, wrap the unlinkSync into another try/catch and swallow and also skip writing the setting. Not having telemetry info for a VSCode session is the lesser evil than not having a functioning extension

davidjenni (changes requested):

src/client/telemetry/localfileusersettings.ts
joem-msft: This file is copied from the PCF implementation in the pac cli repo. This is the kind of divergent change that doesn't scale when we should be using a robust shared implementation (i.e. module).

I'll add the try/catch though.

joem-msft: and swallowing errors here is exactly the problem that I got the BatchedTelemetry code to stop doing so that we would actually be able to know that there are failures happening somewhere. In the end, we were able to fix the root cause of the exceptions in BT. that being said, the try/catch around this delete call is indeed in there. ;-)

src/client/lib/CliAcquisition.ts
joem-msft: fair enough. I'll cleanup the other event below then too.

davidjenni (approved):

PR #56, opened 2021-06-05
Support autocomplete for unsaved files

ashishchoudhary001:
At the moment Autocomplete reads files from the disk and hence is not able to support completion on files under edit.
Now we have stopped reading files from disk and instead have added a document listener to fetch live content from the file being edited.


ameyaapte1 (approved):

src/server/lib/LineReader.ts
```diff
@@ -3,24 +3,17 @@
...
-    let line = liner.next();
-    let lineNumber = 0;
-    let userEditedLine = '';
-
-    while (line) {
-        if (lineNumber == rowIndex) {
-            userEditedLine = line.toString('ascii');
-            break;
+export function getEditedLineContent(rowIndex: number, textDocument: TextDocument): string {
+    const lines = textDocument.getText().split(/\r?\n/g);
```
davidjenni, line 11: since textDoc is now assigned event driven, it's possible for it to be undefined -> use optional chaining .?  to protect here and the for loop in line 13

davidjenni: consider changing the func sig to: textDocument: TextDocument | undefined to let the TS compiler type assert properly

davidjenni (approved):

PR #87, opened 2021-07-27
Disable Remote/Codespaces installation

tehcrashxor:
Disable the extension from running remotely / in Codespaces, as work to support such has not yet been done.

[AB#2373053](https://dev.azure.com/dynamicscrm/1fb98997-2f9e-4734-be8a-9728680447c2/_workitems/edit/2373053) Power Platform VS Code extension gets installed in codespaces even when it is not supported

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,9 @@
   "engines": {
     "vscode": "^1.51.0"
   },
+  "extensionKind": [
+    "ui"
+  ],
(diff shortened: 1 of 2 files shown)


davidjenni (approved):
great, thx! I'll cherry this one, together with my PR #89, to create and ship a refresh

PR #107, opened 2021-09-08
Update Environments Label and multiline tooltips

tehcrashxor:


src/client/lib/PanelComponents.ts
```diff
@@ -59,26 +59,30 @@ export class AuthProfileTreeItem extends vscode.TreeItem {
         }
     }
     private static createTooltip(profile: AuthProfileListing): string {
-        let tooltip = `${profile.Kind}: `;
+        let tooltip = `Profile Kind: ${profile.Kind}: `;
```
davidjenni, line 62: TS idiomatics nit: I'd probably push the tooltip line fragments into an array ```tooltips = [];```, and at the end, do a join:
```return tooltips.join('\n');```

davidjenni (approved):

src/client/lib/PanelComponents.ts
tehcrashxor: updated

PR #146, opened 2022-01-06
Combined Environment and Solutions Panel

swapnilbhalgat:
Combined panel for environment and solutions.
Add context menus to the combined panel.
Remove individual environment and solution panels.


package.json
```diff
@@ -423,18 +409,8 @@
         },
         {
           "id": "pacCLI.envAndSolutionsPanel",
-          "name": "Environments & Solutions",
+          "name": "%pacCLI.envAndSolutionsPanel.title%",
           "when": "config.powerplatform.experimental.activityBarPanels"
```
tehcrashxor, line 413: miscapitalization - should be `...powerPlatform...`

tehcrashxor (changes requested):
Other than the one comment on the miscapitalization (which breaks the display logic), everything else looks good.

tehcrashxor (approved):

PR #167, opened 2022-03-07
powerplatform-vscode,loc checkin for WO61876(3/7/2022)

landy622:
loc checkin for pp-vscode.
Have run "npm run ci" to generate related files for xlf files


v-ly-microsoft:
checkin with the wrong private account. Will checkin with MS account.

PR #169, opened 2022-03-10
updated sample screenshot in readme

davidjenni:


tehcrashxor (commented):
The previous screenshot shows the results of just running `pac` in the screenshot.  Do we want that here as well?

davidjenni:
> The previous screenshot shows the results of just running `pac` in the screenshot. Do we want that here as well?

With the activity pane now visible, I wanted to de-emphasis the terminal and pac CLI somewhat, just to reduce clutter and draw attention to the PP activity pane. I'm fine to add a short pac CLI command, e.g. just the incantaiton but without the console spew??? @kkanakas ?

tehcrashxor (approved):

davidjenni:
merging for now; we can further iterate on the screenshot next week

CC: @kkanakas

PR #202, opened 2022-05-25
Initial changes for VSCode.dev WebExtension for Portals

ashishchoudhary001:
Updating the existing extension to serve as both desktop + web extension

Guidelines for the changes: https://code.visualstudio.com/api/extension-guides/web-extensions#update-existing-extensions-to-web-extensions

dist file/folder structure:

image

fcb for webExtension:

image

disabling commands in virtualWorkspace:

https://code.visualstudio.com/api/extension-guides/virtual-workspaces


ameyaapte1 (approved):

gulpfile.js
```diff
@@ -23,7 +23,7 @@ const log = require('fancy-log');
 const path = require('path');
 const pslist = require('ps-list');
 
-const webPackConfig = require('./webpack.config');
+const [config,webPackConfig] = require('./webpack.config');
```
ameyaapte1, line 26: Can we name this as webConfig?

package.json
```diff
@@ -58,6 +62,7 @@
   ],
   "activationEvents": [
     "onCommand:pacCLI.openDocumentation",
+    "onCommand:microsoft-powerapps-portals.webExtension.init",
```
davidjenni, line 65: suggestion: until the webExt/VSCodeDev functionality reaches MVP/pubPreview quality and feature set, can we keep this under "feature flags", i.e. make it an opt-in feature for the std VSCode Desktop, WSL and GH Codespaces experiences?
In the past we used the regular extensions settings to make this an opt-in experience for the time being

CC: @tehcrashxor

gulpfile.js
ashishchoudhary001: done!

package.json
ashishchoudhary001: Added fcb.

webpack.config.js
```diff
@@ -51,6 +52,58 @@ const config = {
...
+			"path": require.resolve("path-browserify"),
+			'assert': require.resolve('assert')
+		}
+	},
+	module: {
+		rules: [{
+			test: /\.ts$/,
+			exclude: /node_modules/,
+			use: [{
+				loader: 'ts-loader'
```
tehcrashxor, line 88: You might need the `vscode-nls-dev/lib/webpack-loader` as we have above in `nodeConfig` to properly load localized strings.

package.json
```diff
@@ -36,6 +39,7 @@
   ],
```
tehcrashxor, line 39: (Outside of the current changes to the file, and GitHub doesn't let me put the comment on the right line because of that...)

We will probably need to update the `extensionKind` to `["workspace", "ui"]` as shown in the ["Preferred extension location" documentation here.](https://code.visualstudio.com/api/advanced-topics/extension-host#preferred-extension-location) 

Note that I believe that this may lead to an unexpectedly degraded experience for users driving a Remote via a local VS Code installation, so we may need to figure out popup messaging in those scenarios to inform the user.

Example scenario:
User has installed our extension in their local VS Code installation, but then attaches to CodeSpace via the `GitHub Codespaces` extension.
Our extension is not auto-installed (or so I believe) on the attached CodeSpaces remote code-server, thus the extension launches in UI / Web only mode.   The PAC CLI would not be installed on the remote, so the PAC integration both by the terminal and as a backend for the left side UI panel would not be present, and the user might well assume that it would be.

In such a scenario, we should be detecting at extension start that we're running in UI only mode, and pop up an Info box to let the user know that they should install on the remote as well for the full experience.

ashishchoudhary001: I think we should leave extensionKind untouched unless we want to change its existing behaviour.

From the same link: https://code.visualstudio.com/api/advanced-topics/extension-host#preferred-extension-location

"If an extension can run on Node.js and in the browser, a Node.js extension host will be selected if available. There's one exception, when the configuration is VS Code for the Web with Codespaces and the extensionKind is set to ui, then the web extension host is preferred over the remote extension host."

If I understand this correctly, our node extension will be the preferred choice at all the places where we have Node.js including 'VS Code for the Web with Codespaces'. The webExtension should be used only when there is neither (local extension host nor remote extension host i.e. in vscode.dev)  and hence should default to 'web extension host'.

Though I am not an expert in this so please let me know if you think we should double check this with the VSCode team.

tehcrashxor: Ah, my assumption as that the extension installation blocking mechanism in the Extensions tab was based on the extensionKind.   If that is not the case, then we should be good there.

package.json
```diff
@@ -235,6 +249,10 @@
         }
       ],
       "commandPalette": [
+        {
+          "command": "microsoft-powerapps-portals.webExtension.init",
+          "when": "isWeb && config.powerPlatform.experimental.enableWebExtensionFeatures"
```
tehcrashxor, line 254: We will also need the to disable the PAC UI panels in `power-platform-activitybar` and every PAC CLI related command from displaying in the Command Pallet when `isWeb`

ashishchoudhary001: Done.

webpack.config.js
ashishchoudhary001: done.

package.json
tehcrashxor: Missed removing `pacCLI.openDocumentation` from the Command Pallet

package.nls.json
```diff
@@ -48,5 +48,7 @@
   "pacCLI.envAndSolutionsPanel.copyUniqueName.title": "Copy Unique Name",
   "pacCLI.envAndSolutionsPanel.copyVersionNumber.title": "Copy Version Number",
 
-  "microsoft-powerapps-portals.preview-show.title": "PowerApps Portal -> Show preview"
+  "microsoft-powerapps-portals.preview-show.title": "PowerApps Portal -> Show preview",
+  "microsoft-powerapps-portals.webExtension.init.title": "Initialize Web Extension",
+  "%microsoft-powerapps-portals.webExtension.limitation": "PAC CLI is not supported in VS Code for the Web"
```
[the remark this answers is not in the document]
ashishchoudhary001: Removed.

package.json
ashishchoudhary001: Oh I purposefully left this because I think documentation for webExtension might go here. But anyways, for now I have disabled this command as well.

package.json
```diff
@@ -298,83 +322,123 @@
         {
           "command": "pacCLI.authPanel.copyUser",
           "when": "never"
+        },
+        {
+          "command": "pacCLI.openDocumentation",
+          "when": "!virtualWorkspace"
```
tehcrashxor, line 328: Pulling down the branch here and testing has shown some odd behavior.  

Using your new "Run Web Extension in VS Code" launch.json option, all of these PAC commands are still enabled.   Despite that debug environment being `isWeb`, it looks like that environment is _not_ `virtualWorkspace`

Launching the web extension in the browser `npx @vscode/test-web --extensionDevelopmentPath="MyLocalPath"` makes an envrionment that is both `isWeb` and `virtualWorkspace`, and the PAC commands are correctly hidden.

This second method is obviously the one that users will hit instead of the debug environment, but it would be helpful if the debug environment was behaving closer to how the actual target environment behaves.

So we may want to switch all these `"when": "!virtualWorkspace"` to ` "when": "!isWeb"` or ` "when": "!isWeb && !virtualWorkspace"`?

tehcrashxor: ...and ignore that suggestion. 

Tried the branch out in CodeSpaces and swapping `!virtualWorkspace` to `!isWeb`, which breaks the CodeSpaces experience as that environment uses a Web frontend to an actual non-virtual workspace.

It would still be nice if the PAC commands were hidden during "Run Web Extension in VS Code" to better match the user experience, but not at the cost of breaking the CodeSpaces scenario.

davidjenni (approved):

PR #218, opened 2022-06-27
initial changes for webExtension telemetry

ashishchoudhary001:
Setting up telemetry for the Portal WebExtension.


src/web/extension.ts
```diff
@@ -4,9 +4,17 @@
...
+import TelemetryReporter from "@vscode/extension-telemetry";
+import { AI_KEY } from '../client/constants';
+import { randomUUID } from "crypto";
+
+let _telemetry: TelemetryReporter;
 
 export function activate(context: vscode.ExtensionContext): void {
     console.log("Activated web extension!"); // sample code for testing the webExtension
+    // setup telemetry
+    _telemetry = new TelemetryReporter(context.extension.id, context.extension.packageJSON.version, AI_KEY);
```
joem-msft, line 16: keep parity with how the desktop ext uses _telemetry.
here, add `context.subscriptions.push(_telemetry);`

src/web/extension.ts
```diff
@@ -4,9 +4,17 @@
...
+    _telemetry.sendTelemetryEvent("Start", {'web.userId': randomUUID()});
 
     context.subscriptions.push(
         vscode.commands.registerCommand(
```
joem-msft, line 20: add a call to `_telemetry.sendTelemetryEvent('StartCommand', {'commandId': '<id here>'});` like is done in the main extension. even if it doesn't do anything yet.

src/web/extension.ts
joem-msft, line 16: please use the AI key for development instance since this extension is not (hopefully) public yet.

joem-msft, line 16: please add the logged in web user's tenantId if VScode web doesn't already track it.

joem-msft (changes requested):

src/web/client/extension.ts
```diff
@@ -4,14 +4,22 @@
...
+import TelemetryReporter from "@vscode/extension-telemetry";
+import { AI_KEY } from '../../client/constants';
+
+let _telemetry: TelemetryReporter;
 
 export function activate(context: vscode.ExtensionContext): void {
     console.log("Activated web extension!"); // sample code for testing the webExtension
+    // setup telemetry
+    _telemetry = new TelemetryReporter(context.extension.id, context.extension.packageJSON.version, AI_KEY);
+    context.subscriptions.push(_telemetry);
```
joem-msft, line 16: you should add back the `Start` and `Stop` session events, like exists in the other VS Code `extension.ts`

joem-msft: Also, the `activated` event is good to send too.

joem-msft (changes requested):

src/web/client/extension.ts
```diff
@@ -4,14 +4,24 @@
...
+    _telemetry.sendTelemetryEvent("Start");
+    _telemetry.sendTelemetryEvent("activated");
```
joem-msft, line 18: this line should be the last line of the function. as it is in the other extensions.ts

joem-msft (approved):

PR #256, opened 2022-08-12
Custom css file edit enable

tyaginidhi:
- This change addressed loading and saving of custom css files that are saved as Dataverse entityType Annotations.
- Improvements to common utility functions.

image

### DIFF
--- a/src/web/client/common/constants.ts
+++ b/src/web/client/common/constants.ts
@@ -29,6 +29,11 @@ export const PROVIDER_ID = "microsoft";
 export const SCOPE_OPTION = "//.default";
 export const BAD_REQUEST = 'BAD_REQUEST';
 
+export enum HttpMethod {
+    PATCH = 'PATCH',
+    GET = 'GET'
+}
+
 export enum VscodeUrlPathEntityName {
     WEBFILES = "webfiles",
     WEBPAGES = "webpages",
--- a/src/web/client/common/constants.ts
+++ b/src/web/client/common/constants.ts
@@ -29,34 +29,60 @@ export const PROVIDER_ID = "microsoft";
 export const SCOPE_OPTION = "//.default";
 export const BAD_REQUEST = 'BAD_REQUEST';
 
+export enum VscodeUrlPathEntityName {
+    WEBFILES = "webfiles",
+    WEBPAGES = "webpages",
+    WEBTEMPLATES = "webtemplates"
+}
+
+export enum SchemaEntityName {
+    WEBFILES = "adx_webfile",
+    WEBPAGES = "adx_webpage",
+    WEBTEMPLATES = "adx_webtemplate"
+}
+
+export enum DataverseUrlPathEntityName {
+    WEBFILES = "annotations",
+    WEBPAGES = "adx_webpages",
+    WEBTEMPLATES = "adx_webtemplate" // TODO - Update to right call name
+}
+
+export enum ExportType {
+    SubFolders = "SubFolders",
+    SingleFolder = "SingleFolder",
+    SingleFile = "SingleFile"
+}
+
 export const columnExtension = new Map([
     ["customcss.css", "adx_customcss"],
     ["customjs.js", "adx_customjavascript"],
     ["webpage.copy.html", "adx_copy"],
     ["adx_customcss", "customcss.css"],
     ["adx_customjavascript", "customjs.js"],
-    ["adx_copy", "webpage.copy.html"]
+    ["adx_copy", "webpage.copy.html"],
+    ["documentbody", "css"]
 ]);
 
 export const entityFolder = new Map([
     ["webpages", "web-pages"],
     ["webtemplates", "web-templates"],
-    ["webfiles", "web-files"],
+    ["annotations", "web-files"],
     ["contentpages", "content-pages"],
     ["adx_webpages", "web-pages"]
 ]);
 
 export const pathParamToSchema = new Map([
-    ["webpages", "adx_webpages"],
-    ["webtemplates", "adx_webtemplates"],
+    [VscodeUrlPathEntityName.WEBPAGES, "adx_webpages"],
+    [VscodeUrlPathEntityName.WEBFILES, "annotations"],
+    [VscodeUrlPathEntityName.WEBTEMPLATES, "adx_webtemplates"],
     ["adx_webpages", "adx_webpage"],
     ["adx_webtemplates", "adx_webtemplate"],
     ["adx_websites", "adx_website"],
     ["adx_portallanguages", "adx_portallanguage"],
     ["adx_websitelanguages", "adx_websitelanguage"],
-    ["webfiles", "adx_webfiles"],
     ["adx_webfiles", "adx_webfile"],
-    ["adx_copy", "adx_copy"]
+    ["adx_copy", "adx_copy"],
+    ["annotations", "adx_webfile"],
 ]);
 
 export enum queryParameters {
--- a/src/web/client/common/constants.ts
+++ b/src/web/client/common/constants.ts
@@ -63,7 +63,7 @@ export enum schemaEntityName {
 export enum dataverseUrlPathEntityName {
     WEBFILES = "annotations",
     WEBPAGES = "adx_webpages",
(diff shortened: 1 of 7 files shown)


ashishchoudhary001 (commented):
Please also test if the existing scenarios (i.e. webpage editing) are working fine since this PR changes some of the existing stuff.

src/web/client/common/remoteFetchProvider.ts
```diff
@@ -4,103 +4,190 @@
...
-        const data = await response.json();
-        if (data.value?.length >= 0) {
-            for (let counter = 0; counter < data.value.length; counter++) {
-                createContentFiles(data[counter], entity, queryParamsMap, entitiesSchemaMap, languageIdCodeMap, portalFs, dataverseOrgUrl, accessToken, entityId, websiteIdToLanguage);
-            }
-        } else {
-            createContentFiles(data, entity, queryParamsMap, entitiesSchemaMap, languageIdCodeMap, portalFs, dataverseOrgUrl, accessToken, entityId, websiteIdToLanguage);
+
+        const result = await response.json();
+        const data = getParameterizedRequestUrlKey(entity, entityId) === Constants.SINGLE_ENTITY_URL_KEY ? [result] : result.value;
```
ashishchoudhary001, line 61: Something seems odd here. How does this work for 'webfiles'?

src/web/client/common/portalSchema.ts
```diff
@@ -766,13 +766,14 @@ export const portal_schema_data = {
...
+                "_primarynamefield": "filename",
                 "_disableplugins": "true",
                 "_foldername": "web-files",
                 "_propextension": "webfile",
                 "_exporttype": "SingleFolder",
-                "_fetchQueryParameters": "?$filter=_objectid_value%20eq%{_objectid_value}%20&$select=documentbody",
-                "_attributes": "documentbody"
+                "_fetchQueryParameters": "?$filter=_objectid_value%20eq%20{entityId}%20&$select=documentbody,filename,annotationid",
+                "_attributes": "documentbody",
+                "_mappingEntityId": "annotationid"
```
ashishchoudhary001, line 776: @vedg please take a note of these changes in the schema. We will need to add these in the newSchema as well.

src/web/client/common/constants.ts
```diff
@@ -29,34 +29,65 @@ export const PROVIDER_ID = "microsoft";
...
+export enum SchemaEntityName {
+    WEBFILES = "adx_webfile",
+    WEBPAGES = "adx_webpage",
+    WEBTEMPLATES = "adx_webtemplate"
+}
+
+export enum DataverseUrlPathEntityName {
+    WEBFILES = "annotations",
+    WEBPAGES = "adx_webpages",
+    WEBTEMPLATES = "adx_webtemplate" // TODO - Update to right call name
```
ashishchoudhary001, line 52: should be adx_webtemplates

src/web/client/common/constants.ts
```diff
@@ -29,34 +29,65 @@ export const PROVIDER_ID = "microsoft";
 export const SCOPE_OPTION = "//.default";
 export const BAD_REQUEST = 'BAD_REQUEST';
 
+export enum HttpMethod {
+    PATCH = 'PATCH',
+    GET = 'GET'
+}
+
+export enum VscodeUrlPathEntityName {
```
ashishchoudhary001, line 37: Please add comments to explain significance of these enums.

e.g.
1. This is the simplified/generic name of the entity passed in the URL by the clients
2. This is the entity name in the schema file. Schema file contains a lot of properties about how to fetch/save the entity and how to display it in the file explorer.
3. These are the OData names of the resources required for fetch and save calls.

src/web/client/utility/UrlBuilder.ts
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ */
+
+import { DataverseUrlPathEntityName, HttpMethod, MULTI_ENTITY_URL_KEY, pathParamToSchema, SINGLE_ENTITY_URL_KEY } from "../common/constants";
+import { dataSourcePropertiesMap } from "../common/localStore";
+
+export const getParameterizedRequestUrlTemplate = (entity: string, entityId: string, isSingleEntity: boolean) => {
+    if (isSingleEntity || (entity === DataverseUrlPathEntityName.WEBPAGES && entityId)) {
```
ashishchoudhary001, line 10: why do we have special condition for WEBPAGES?

src/web/client/common/remoteFetchProvider.ts
ashishchoudhary001, line 61: This should be changed. We should structure our url in such a way that we always get our data from "result.value". Let's discuss about this.

src/web/client/utility/CommonUtility.ts
```diff
@@ -0,0 +1,33 @@
+/*
```
vedg, line 1: Please follow filename convention of this repo

src/web/client/common/remoteFetchProvider.ts
tyaginidhi: For single entity query the results are returned as flat list of parameters for that single entity. But in all other cases the entities (or even if its a single object returned as part of filter query) are returned inside "value" parameter as a array of objects. So, for webfiles, and multi-files the array of object for iteration will be result.value. As an improvement to this code we should define the response object in case of multi/webfile and single file fetch here for clarity.

src/web/client/common/portalSchema.ts
tyaginidhi: We should also discuss some changes to existing schema. Currently, we convert entityType name sent in vscode.dev url from portal to another entity type, which is used for making the dataverse call. And then we further map it to the enityType/Name we maintain in the schema. I believe we dont need to maintain different dataverse call names v/s schema names. Or, is there some background i am missing here. 
This simple change will simplify our pathParamToEntity conversion logic. Currently, we are using the same mapping object for both of the above conversions. Either we simplify the logic above, or atleast create two separate mapping objects.

src/web/client/common/constants.ts
tyaginidhi: These names should be self explanatory. We should update these, if they are not clear. Will add the comments as well here.

src/web/client/utility/UrlBuilder.ts
tyaginidhi: Good catch, will simplify this.

src/web/client/common/remoteFetchProvider.ts
tyaginidhi: Lets go over this PR in our sync once.

src/web/client/common/portalSchema.ts
```diff
@@ -766,13 +766,14 @@ export const portal_schema_data = {
                 "_displayname": "Web File",
                 "_etc": "10020",
                 "_primaryidfield": "adx_webfileid",
-                "_primarynamefield": "adx_name",
+                "_primarynamefield": "filename",
```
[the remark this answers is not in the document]
tyaginidhi: Lets compare the dataverse call here then. I believe that is the gap here.

src/web/client/common/portalSchema.ts
tyaginidhi: TODO: Leave a comment in code for this logic

src/web/client/common/constants.ts
tyaginidhi: TODO: Put detailed comment on logic

src/web/client/common/portalSchema.ts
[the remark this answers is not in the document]
tyaginidhi: TODO: Verify from studio again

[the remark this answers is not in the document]
tyaginidhi: Studio has this mapping:
export class AnnotationAttributes {
  public static readonly FILE_NAME = 'fileName';
  public static readonly DOCUMENT_BODY = 'documentBody';
  public static readonly FILE_SIZE = 'fileSize';...

ashishchoudhary001 (approved):

PR #272, opened 2022-08-20
Site visibility dialog

vedg:
Added site visibility dialog

updated the image with new changes 

image

### DIFF
--- a/src/web/client/extension.ts
+++ b/src/web/client/extension.ts
@@ -48,10 +48,10 @@ export function activate(context: vscode.ExtensionContext): void {
                 }
 
                 if (queryParamsMap.get(SITE_VISIBILITY) === PUBLIC) {
-                    const edit: vscode.MessageItem = { isCloseAffordance: true, title: "Edit the site" };
-                    const siteMessage = "Be careful making changes. Anyone can see changes you make right away. To edit in private, go to Set up to change the site visibility to private";
+                    const edit: vscode.MessageItem = { isCloseAffordance: true, title: localize("microsoft-powerapps-portals.webExtension.init.sitevisibility.edit","Edit the site") };
+                    const siteMessage = localize("microsoft-powerapps-portals.webExtension.init.sitevisibility.edit.desc","Be careful making changes. Anyone can see the changes you make immediately. Choose Edit the site to make edits, or close the editor tab to cancel without editing.");
                     const options = { detail: siteMessage, modal: true };
-                    vscode.window.showWarningMessage("You are editing a live, public site ", options, edit);
+                    await vscode.window.showWarningMessage(localize("microsoft-powerapps-portals.webExtension.init.sitevisibility.edit.title","You are editing a live, public site "), options, edit);
                 }
 
                 let accessToken: string;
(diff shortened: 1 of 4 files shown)


davidjenni (changes requested):
Popping up a modal dialog as part of the extension's activation is too intrusive to users, even here where it is on the vscode.dev path. Please defer showing that warning (even if modal) when the user actually does an up/download.

neerajnandwana-msft:
> Popping up a modal dialog as part of the extension's activation is too intrusive to users, even here where it is on the vscode.dev path. Please defer showing that warning (even if modal) when the user actually does an up/download.

These changes are part of CELA feedback, this requirement is same as what we have discussed around PAC CLI information message.

neerajnandwana-msft:
> Popping up a modal dialog as part of the extension's activation is too intrusive to users, even here where it is on the vscode.dev path. Please defer showing that warning (even if modal) when the user actually does an up/download.

These changes are part of CELA feedback, and it is same as what we have discussed around PAC CLI information message.

src/web/client/extension.ts
```diff
@@ -34,7 +34,8 @@ export function activate(context: vscode.ExtensionContext): void {
             "microsoft-powerapps-portals.webExtension.init",
             async (args) => {
                 _telemetry.sendTelemetryEvent("StartCommand", { 'commandId': 'microsoft-powerapps-portals.webExtension.init' });
-                vscode.window.showInformationMessage(localize("microsoft-powerapps-portals.webExtension.init.message", "Opening VS Code for the web ..."));
+
+                vscode.window.showInformationMessage("Opening VS Code for the web ...");
```
vedg, line 38: will update this in next pr

src/web/client/extension.ts
```diff
@@ -45,6 +46,14 @@ export function activate(context: vscode.ExtensionContext): void {
                         queryParamsMap.set(pair[0], pair[1]);
                     }
                 }
+
+                if (queryParamsMap.get(SITE_VISIBILITY) === PUBLIC) {
+                    const edit: vscode.MessageItem = { isCloseAffordance: true, title: localize("microsoft-powerapps-portals.webExtension.init.sitevisibility.edit","Edit the site") };
```
ashishchoudhary001, line 51: Please confirm if we need: isCloseAffordance

webpack.config.js
```diff
@@ -85,13 +85,6 @@ const webConfig = {
 			test: /\.ts$/,
 			exclude: /node_modules/,
 			use: [{
-                // vscode-nls-dev loader:
```
ashishchoudhary001, line 88: Please put the tracking VSCode dev localization bug link. For localization eventually we will have to bring this back.

ashishchoudhary001 (approved):

webpack.config.js
vedg: https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/2896662

src/web/client/extension.ts
vedg: yes this is needed for default close and Esc button behavior

PR #283, opened 2022-09-08
Support Web template and content snippet 

tyaginidhi:
Web template load in studio:
image

image

Content snippet load in studio:
image

image


src/web/client/common/portalSchema.ts
```diff
@@ -1311,9 +1311,11 @@ export const portal_schema_data = {
                 "_disableplugins": "true",
                 "_foldername": "content-snippets",
                 "_propextension": "contentsnippet",
-                "_exporttype": "SubFolders",
+                "_exporttype": "SingleFolder",
```
ashishchoudhary001, line 1314: @tyaginidhi I think this change we will have to revert. Let's discuss about this. I think we are missing language information for the content snippets?

tyaginidhi: Lets make the content snippet language change as part of next PR specific to content snippet

ashishchoudhary001 (approved):

PR #284, opened 2022-09-08
Walkthrough for VSCode Web

ashishchoudhary001:
Walkthrough for VSCode Web:

This walkthrough only opens in Web and at the moment it is not available for the Desktop/Codespaces.

This is controlled by the attribute: "when": "isWeb && virtualWorkspace"

image


package.json
```diff
@@ -64,6 +64,11 @@
   "activationEvents": [
     "onCommand:pacCLI.openDocumentation",
     "onCommand:microsoft-powerapps-portals.webExtension.init",
+    "onCommand:powerplatform-walkthrough.overview-learn-more",
```
tehcrashxor, line 67: It looks like you're referencing these commands here, and registering the handlers in the extension.ts, but never actually declaring the commands in the `contributes -> commands` section of the package.json

ashishchoudhary001: Yes these are kind of internal commands and are invoked by clicking on links/buttons described in the package.nls.json file.
E.g. [Learn More](command:powerplatform-walkthrough.overview-learn-more)

Please check the official VSCode samples: https://github.com/microsoft/vscode-extension-samples/blob/main/getting-started-sample/package.json

tehcrashxor: Interesting.   I assume that all commands needed to be listed in the Contributes section, but it looks like that's only necessary when the command needs a title or icon, as all of the commands we've previously made have but that these do not.

Double checked by running it locally, and it all appears to work fine.

tehcrashxor (approved):

PR #285, opened 2022-09-09
Updated forward link and Walkthrough-Overview text

ashishchoudhary001:
Updated forward link and Overview text.

image


package.nls.json
```diff
@@ -54,7 +54,7 @@
   "microsoft-powerapps-portals.walkthrough.title": "Edit Power Pages code",
   "microsoft-powerapps-portals.walkthrough.description": "Discover how you can make complete, modern websites right in VS Code",
   "microsoft-powerapps-portals.walkthrough.overview.title": "Overview",
-  "microsoft-powerapps-portals.walkthrough.overview.description": "Power Pages extension enables a customized file system which is synced with the project structure of the Power Pages studio.\n \n The preview is linked to Power Pages and enables you to view your changes.\n[Learn More](command:powerplatform-walkthrough.overview-learn-more)",
+  "microsoft-powerapps-portals.walkthrough.overview.description": "Power Pages extension enables a customized file system which is synced with the project structure of the Power Pages studio.\n[Learn More](command:powerplatform-walkthrough.overview-learn-more)",
```
tehcrashxor, line 57: Didn't notice this earlier, but you should use the format above (example - `pacCLI.authPanel.welcome.whenInteractiveNotSupported`'s entry) to add translation comments, so that the translators in the Loc team know to keep the exact format needed for the markdown portions like `[Learn More](command:powerplatform-walkthrough.overview-learn-more)`, as they've occasionally used different bracket styles more common to other languages.

tehcrashxor (approved):

tyaginidhi (approved):

package.nls.json
tehcrashxor: #292 is a good example of this - Loc team didn't have instructions to keep the markdown format exactly the same, so we ended up with a bunch of translations with spaces breaking the link format (eg, `[translated text] (command)`), which causes those links (or buttons in the case of some views) to be displayed as string literals instead of as the link/button we expect.

ashishchoudhary001: Yup hadn't noticed this pattern unless you pointed out. Have fixed it now, please review.

tehcrashxor: Yep, looks good.

tehcrashxor (approved):

PR #288, opened 2022-09-12
powerplatform-vscode Loc checkin for WO71206

v-hifuku-msft:
powerplatform-vscode Loc checkin for WO71206


loc/translations-import/vscode-powerplatform.fr.xlf
```diff
@@ -145,6 +145,46 @@ L’authentification interactive n’est pas disponible pour les scénarios dist
         <source xml:lang="en">PAC CLI is not supported in VS Code for the Web</source>
         <target state="translated">PAC CLI n’est pas pris en charge dans VS Code pour le Web</target>
       </trans-unit>
+      <trans-unit id="microsoft-powerapps-portals.walkthrough.title">
```
tehcrashxor, line 148: Looks like there are a decent number of strings in the `package.json` section of the translated xlf files here that don't have translations added.

I _think_ that should be fine, as these are used by the Web Only version of the extension, which currently doesn't load translations anyway due to a VS Code Web issue.   @ashishchoudhary001 , can you confirm that?

tehcrashxor (approved):
Spot checked a couple languages, and the strings used by the main version of the extension are loading as expected.   Checking with @ashishchoudhary001 about the missing translations used by the Web extension per the other comment.

v-hifuku-msft:
Please note that additional stings were added later last week. We are handing them to localization today. ETA is this Thursday.

Thanks,
Hiro

loc/translations-import/vscode-powerplatform.fr.xlf
tehcrashxor: Resolving comment per Hiro's reply below - these strings will be localized later this week.

PR #299, opened 2022-09-28
changes to show walkthrough only for the first run experience

ashishchoudhary001:
Contains changes to show walkthrough only for the first run experience

image

### DIFF
--- a/src/web/client/common/remoteFetchProvider.ts
+++ b/src/web/client/common/remoteFetchProvider.ts
@@ -158,7 +158,7 @@ async function createContentFiles(
         PowerPlatformExtensionContextManager.updateSingleFileUrisInContext(vscode.Uri.parse(fileUri));
 
         // Display only the last file
-        vscode.window.showTextDocument(vscode.Uri.parse(fileUri));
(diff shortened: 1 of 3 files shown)


src/web/client/common/remoteFetchProvider.ts
```diff
@@ -158,7 +158,7 @@ async function createContentFiles(
         PowerPlatformExtensionContextManager.updateSingleFileUrisInContext(vscode.Uri.parse(fileUri));
 
         // Display only the last file
-        vscode.window.showTextDocument(vscode.Uri.parse(fileUri));
+        vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(fileUri), {background: true, preview: false});
```
tyaginidhi, line 161: this will still prompt to open the default file in editor right?

tyaginidhi (approved):

src/web/client/common/remoteFetchProvider.ts
ashishchoudhary001: Prompt?
But yes, it will open the default file in editor just as it used to do earlier (only difference is this time it will open it in non-preview[Permanent] mode).

PR #317, opened 2022-10-18
additional telemetry for webExtension

ashishchoudhary001:


src/web/client/common/authenticationProvider.ts
```diff
@@ -31,6 +31,7 @@ export async function dataverseAuthentication(dataverseOrgURL: string): Promise<
                 sendErrorTelemetry(telemetryEventNames.WEB_EXTENSION_NO_ACCESS_TOKEN);
             }
         }
+        sendInfoTelemetry("Auth", { 'web.userId': session.id });
```
ashishchoudhary001, line 34: check about data tagging/any other compliance before merging the PR.

src/web/client/telemetry/webExtensionTelemetry.ts
```diff
@@ -122,32 +134,35 @@ export function sendInfoTelemetry(eventName: string, properties?: Record<string,
...
         },
         measurements: {
             durationInMillis: (duration) ? duration : 0
         }
     }
     if (errorMessage) {
-        const errorMessages: string[] = [];
-        errorMessages.push(errorMessage);
-        _telemetry?.sendTelemetryErrorEvent(telemetryData.eventName, telemetryData.properties, telemetryData.measurements, errorMessages);
+        // TODO: test these events before merging the PR
```
ashishchoudhary001, line 151: Verify the telemetry data on the Preview Build before merging.

src/web/client/telemetry/webExtensionTelemetry.ts
```diff
@@ -122,32 +134,35 @@ export function sendInfoTelemetry(eventName: string, properties?: Record<string,
...
+        const error: Error = new Error(errorMessage);
+        _telemetry?.sendTelemetryException(error, telemetryData.properties, telemetryData.measurements);
     } else {
-        _telemetry?.sendTelemetryErrorEvent(telemetryData.eventName, telemetryData.properties, telemetryData.measurements);
+        // TODO: test these events before merging the PR
+        _telemetry?.sendTelemetryEvent(telemetryData.eventName, telemetryData.properties, telemetryData.measurements);
```
ashishchoudhary001, line 156: Verify the telemetry data on the Preview Build before merging.

src/web/client/common/authenticationProvider.ts
tyaginidhi: @tehcrashxor We are planning to log userId as part of this telemetry change. Any GDPR concerns here? We already log this id in other code paths, and this is compliant, correct?

ashishchoudhary001: @tyaginidhi I think we should replace session.id(identifier of the session) with session.account.id(unique identifier of the account). Can you once test locally to check what is the value for this...must be a GUID but is it encoded/hashed.
Maybe you can verify it using: https://unify.services.dynamics.com/CRM/Org/<orgId>/Organization/ApplicationUsers

For compliance we will have to check with @tehcrashxor

tyaginidhi: We should not use any of the session field in logging. Verified on my local that all of this is PII:
image

tyaginidhi (approved):

src/web/client/common/remoteSaveProvider.ts
```diff
@@ -32,32 +34,32 @@ export async function saveData(
...
+                sendAPIFailureTelemetry(requestUrl, entityName, httpMethod.PATCH, new Date().getTime() - requestSentAtTime, response.json.toString());
                 showErrorDialog(localize("microsoft-powerapps-portals.webExtension.backend.error", "There’s a problem on the back end"), localize("microsoft-powerapps-portals.webExtension.retry.desc", "Try again"));
                 throw new Error(response.statusText);
             }
 
-            sendAPISuccessTelemetry(requestUrl, new Date().getTime() - requestSentAtTime);
+            sendAPISuccessTelemetry(requestUrl, entityName, httpMethod.PATCH, new Date().getTime() - requestSentAtTime);
         }
         catch (error) {
             const authError = (error as Error)?.message;
```
amitjoshi438, line 61: Is this a authError?

amitjoshi438 (approved):

src/web/client/common/remoteFetchProvider.ts
```diff
@@ -75,7 +75,7 @@ export async function fetchDataFromDataverseAndUpdateVFS(
         else {
```
joem-msft, line 75: This condition will only work for English locales. You should probably update your code to detect specific statusCodes first. Or handle specific error types.

You would also likely benefit from using `node-fetch` instead so that you have guaranteed Error-derived exceptions vs strings being thrown. See: [ERROR-HANDLING.md](https://github.com/node-fetch/node-fetch/blob/main/docs/ERROR-HANDLING.md)

src/web/client/common/remoteSaveProvider.ts
```diff
@@ -32,32 +34,32 @@ export async function saveData(
         }
         requestBody = JSON.stringify(data);
     } else {
-        sendAPIFailureTelemetry(requestUrl, 0, BAD_REQUEST); // no API request is made in this case since we do not know in which column should we save the value
+        sendAPIFailureTelemetry(requestUrl, entityName, httpMethod.PATCH, 0, BAD_REQUEST); // no API request is made in this case since we do not know in which column should we save the value
```
joem-msft, line 37: I would consider NOT using the sendAPIFailureTelemetry event for this kind of error since there's no actual API being called yet.
Record this simply as another error or perhaps throw so the caller knows this is an invalid input?
this seems more like a caller having invalid parameters vs an API issue.

src/web/client/extension.ts
```diff
@@ -22,7 +22,6 @@ export function activate(context: vscode.ExtensionContext): void {
     context.subscriptions.push(_telemetry);
 
     setTelemetryReporter(_telemetry);
-    sendInfoTelemetry("Start");
```
joem-msft, line 25: add the Start event back. This is required for our telemetry pipeline to work. 'Start' means 'session started'.
You should add on the user's pac CLI user id, as is done in the desktop client's extension.

PR #320, opened 2022-10-20
2915250 - Content snippet language Id changes in fileName

tyaginidhi:
image


src/web/client/common/constants.ts
```diff
@@ -17,14 +17,15 @@ export const CONTENT_PAGES = 'contentpages';
 export const SCHEMA = "schema";
 export const WEBSITE_ID = "websiteId";
 export const WEBSITE_NAME = "websiteName";
-export const DEFAULT_LANGUAGE_CODE = ' ';
+export const DEFAULT_LANGUAGE_CODE = 'en-US';
```
ashishchoudhary001, line 20: I think let's not change this, it was purposefully kept empty.
We can have orgs with non-english base language (e.g. Japanese) and these orgs may not have english contentpages/snippets at all. Though '' doesn't really show an error but at least it doesn't misguide the user.

ashishchoudhary001 (approved):
All other changes look good.

PR #328, opened 2022-10-21
Adding codeTour and Swagger docs

ashishchoudhary001:
Adding codeTour and Swagger docs. Going ahead more code tours can be added in the .tours folder.


tyaginidhi:
Being taken up as a different PR and code tour is already part of the walkthrough

PR #336, opened 2022-11-01
Use VS Code's Telemetry setting for the PAC backend

tehcrashxor:
AB#3012127 - VS Code Extension's PAC backend should honor the VS Code Telemetry Disable setting

When VS Code is set to disable telemetry, our extension already picks up that state and disables extension-specific telemetry.
However, the copy of PAC running as a backend for the extension's UI uses PAC's own telemetry enable/disable system, and prior to this change would not follow VS Code's setting.

### DIFF
--- a/src/client/pac/PacWrapper.ts
+++ b/src/client/pac/PacWrapper.ts
@@ -70,6 +70,11 @@ export class PacInterop implements IPacInterop {
 
             const env : NodeJS.ProcessEnv = {...process.env, 'PP_TOOLS_AUTOMATION_AGENT': this.context.automationAgent };
 
+            // If the VS Code telemetry is disabled, disable telemetry on the PAC backing the Extension's UI
+            if (!vscode.env.isTelemetryEnabled) {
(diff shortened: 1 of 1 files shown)


tyaginidhi (approved):

davidjenni (approved):

src/client/pac/PacWrapper.ts
```diff
@@ -70,6 +70,11 @@ export class PacInterop implements IPacInterop {
 
             const env : NodeJS.ProcessEnv = {...process.env, 'PP_TOOLS_AUTOMATION_AGENT': this.context.automationAgent };
 
+            // If the VS Code telemetry is disabled, disable telemetry on the PAC backing the Extension's UI
+            if (!vscode.env.isTelemetryEnabled) {
+                env['PP_TOOLS_TELEMETRY_OPTOUT'] = 'true';
```
joem-msft, line 75: Love this. it makes it so we don't also turn off telemetry for other installations of Pac CLI on the same machine.

joem-msft (approved):

PR #337, opened 2022-11-01
Build Preview package as part of official build

tehcrashxor:
Build both the Standard and Preview .vsix packages

[AB#3010171](https://dev.azure.com/dynamicscrm/1fb98997-2f9e-4734-be8a-9728680447c2/_workitems/edit/3010171)


petrochuk (approved):

Preview.md
```diff
@@ -0,0 +1,3 @@
+# Power Platform Tools [PREVIEW]
+
```
joem-msft, line 2: Should this be named: README.Preview.md?
It's clear this file applies to the preview build only. but imagine if 'Preview' wasn't there, what would it be called?

gulpfile.js
```diff
@@ -47,17 +47,6 @@ async function clean() {
     return fs.emptyDir(distdir);
 }
 
-function setTelemetryTarget() {
```
joem-msft, line 50: Please bring these files back as this pattern of using different files that are code is consistent with how we manage this stuff in our other NodeJS products. The fact that they're still Dev vs Prod configuration still applies in the new *-Preview deployment model, it's just that the Preview ext will currently always pick the Dev configuration.
In the near future when we try to support EUDB regions, I only want to modify these code files, and they'll reuse code we've creted elsewhere so we don't reinvent the wheel.

gulpfile.js
```diff
@@ -215,10 +204,46 @@ const test = gulp.series(testUnitTests, testWeb);
...
     fs.emptyDirSync(packagedir);
-    return vsce.createVSIX({
+
+    // Set Preview extension name
+    await npm(['pkg', 'set', 'name=powerplatform-vscode-preview']);
+    await npm(['pkg', 'set', 'displayName="Power Platform Tools [PREVIEW]"']);
+    await npm(['pkg', 'set', 'description="Unsupported extension for testing Power Platform Tools"']);
+    await npm(['pkg', 'set', `aiKey=${devAiKey}`]);
+
+    gulp.src('README.md')
```
joem-msft, line 218: all this file copying and renaming is going to be problematic.
What happens when errors occur during these tasks?

And how does the local developer build things? Will it dirty up their local git changes? since the readme.md now has a different content?

gulpfile.js
```diff
@@ -215,10 +204,46 @@ const test = gulp.series(testUnitTests, testWeb);
...
+        preRelease: true,
+    });
+
+    // Reset to default name for standard package
+    await npm(['pkg', 'set', 'name=powerplatform-vscode']);
+    await npm(['pkg', 'set', 'displayName="Power Platform Tools"']);
+    await npm(['pkg', 'set', 'description="Tooling to create Power Platform solutions & packages, manage Power Platform environments and edit Power Apps Portals"']);
+    await npm(['pkg', 'set', `aiKey=${isOfficialBuild ? prodAiKey : devAiKey}`]);
+
+    gulp.src('README.original.md')
```
joem-msft, line 236: put this in a catch block?

src/client/extension.ts
```diff
@@ -34,7 +33,7 @@ export async function activate(
     _context = context;
 
     // setup telemetry
-    _telemetry = new TelemetryReporter(context.extension.id, context.extension.packageJSON.version, AI_KEY);
+    _telemetry = new TelemetryReporter(context.extension.id, context.extension.packageJSON.version, context.extension.packageJSON.aiKey);
```
joem-msft, line 36: please undo this change. It diverges from how we expect to model selecting the AI resource that belongs to the region configured by the user.

joem-msft (changes requested):

gulpfile.js
tehcrashxor: The point of setting these _at package time_ was to prevent the need to build twice.  Looking at the official builds though, the build+package step is only 45 seconds, so it should be fine to go the build-twice route.

Preview.md
tehcrashxor: renamed

gulpfile.js
tehcrashxor: Switched to packaging a single VSIX in the package step, and changed the Official Build to build twice to generate both packages

gulpfile.js
tehcrashxor: Original readme is reset after the preview package is done, and the .gitignore update prevents "README.original.md" from polluting the git diffs

README.Preview.md
```diff
@@ -0,0 +1,3 @@
+# Power Platform Tools [PREVIEW]
```
davidjenni, line 1: I'd prefer a single readme.md, where the gulpfile just injects the "scary text" right after this header line

gulpfile.js
```diff
@@ -215,10 +217,36 @@ const test = gulp.series(testUnitTests, testWeb);
...
 async function packageVsix() {
-    fs.emptyDirSync(packagedir);
-    return vsce.createVSIX({
+    if (isPreviewBuild) {
+        // Set Preview extension settings
+        await npm(['pkg', 'set', 'name=powerplatform-vscode-preview']);
+        await npm(['pkg', 'set', 'displayName="Power Platform Tools [PREVIEW]"']);
+        await npm(['pkg', 'set', 'description="Unsupported extension for testing Power Platform Tools"']);
+
+        gulp.src('README.md')
```
davidjenni, line 226: see above comment: single readme.md; if isPreviewBuild, inject the warning text

README.Preview.md
davidjenni: also, use MD tagging to make this text more intrusive to the reader, e.g. bold, or header2 
Doesn't have to look "pretty", needs to pop in users' eyes

davidjenni (changes requested):

README.Preview.md
tehcrashxor: Switched to single readme file via text replacement, and changed warning text to H2

gulpfile.js
tehcrashxor: Switched to single readme file via text replacement, and changed warning text to H2

gulpfile.js
tehcrashxor: Off of David's idea, swapped to single readme file with text replacement

gulpfile.js
```diff
@@ -215,10 +217,30 @@ const test = gulp.series(testUnitTests, testWeb);
...
+    const stanardHeader = '# Power Platform Extension';
+    const previewHeader = '# Power Platform Tools [PREVIEW]\n\n## This extension is used for internal testing against targets such as vscode.dev which require Marketplace published extensions, and is not supported.';
+
+    const setPackageInfo = async function(setPreviewState) {
+        await npm(['pkg', 'set', `name=${setPreviewState ? 'powerplatform-vscode-preview' : 'powerplatform-vscode'}`]);
+        await npm(['pkg', 'set', `displayName=${setPreviewState ? '"Power Platform Tools [PREVIEW]"' : '"Power Platform Tools"'}`]);
+        await npm(['pkg', 'set', `description=${setPreviewState ? '"Unsupported extension for testing Power Platform Tools"' : '"Tooling to create Power Platform solutions & packages, manage Power Platform environments and edit Power Apps Portals"'}`]);
+
+        gulp.src('README.md')
+            .pipe(setPreviewState ? replace(stanardHeader, previewHeader) : replace(previewHeader, stanardHeader))
```
davidjenni, line 229: typo: stan>d<ardHeader

gulpfile.js
```diff
@@ -215,10 +217,30 @@ const test = gulp.series(testUnitTests, testWeb);
 const testInt = gulp.series(testDebugger);
 
 async function packageVsix() {
-    fs.emptyDirSync(packagedir);
-    return vsce.createVSIX({
+    const stanardHeader = '# Power Platform Extension';
+    const previewHeader = '# Power Platform Tools [PREVIEW]\n\n## This extension is used for internal testing against targets such as vscode.dev which require Marketplace published extensions, and is not supported.';
```
davidjenni, line 221: fine to have this text inline for now; but should we decide to add more disclaimer text, we'd need to factor the text out into a disclaimer.md

davidjenni (approved):

gulpfile.js
tehcrashxor: function rewritten

gulpfile.js
```diff
@@ -215,10 +217,30 @@ const test = gulp.series(testUnitTests, testWeb);
 const testInt = gulp.series(testDebugger);
 
 async function packageVsix() {
-    fs.emptyDirSync(packagedir);
-    return vsce.createVSIX({
+    const setPackageInfo = async function(setPreviewState) {
+        const standardHeader = '# Power Platform Extension';
```
joem-msft, line 221: nit: Since each config has its own settings, it may be more scalable to create a json object with each option in a property.

```js
const pkgOptions = {
   name: 'powerplatform-vscode',
   displayName: 'Power Platform Tools',
   description: 'Tooling to create Power Platform solutions & packages, manage Power Platform environments and edit Power Apps Portals',
   readmeHeader: '# Power Platform Extension',
};
if(setPreviewState) {
   pkgOptions = {
      name: 'powerplatform-vscode-preview',
      displayName: 'Power Platform Tools [PREVIEW]',
      description: 'Unsupported extension for testing Power Platform Tools',
      readmeHeader: '# Power Platform Tools [PREVIEW]\n\n## This extension is used for internal testing against targets such as vscode.dev which require Marketplace published extensions, and is not supported.',
   };
}
```

Then, when applying the options, you can focus on encoding without complicating with conditionals:
```js
await npm(['pkg', 'set', `name=${pkgOptions.name}`]);
await npm(['pkg', 'set', `displayName="${pkgOptions.displayName}"`]);
await npm(['pkg', 'set', `description="${pkgOptions.description}"'}`]);

gulp.src('README.md')
   .pipe(replace(standardHeader, pkgOptions.readmeHeader)
   .pipe(gulp.dest('./'));
```

gulpfile.js
```diff
@@ -215,10 +217,30 @@ const test = gulp.series(testUnitTests, testWeb);
...
+        const previewHeader = '# Power Platform Tools [PREVIEW]\n\n## This extension is used for internal testing against targets such as vscode.dev which require Marketplace published extensions, and is not supported.';
+
+        await npm(['pkg', 'set', `name=${setPreviewState ? 'powerplatform-vscode-preview' : 'powerplatform-vscode'}`]);
+        await npm(['pkg', 'set', `displayName=${setPreviewState ? '"Power Platform Tools [PREVIEW]"' : '"Power Platform Tools"'}`]);
+        await npm(['pkg', 'set', `description=${setPreviewState ? '"Unsupported extension for testing Power Platform Tools"' : '"Tooling to create Power Platform solutions & packages, manage Power Platform environments and edit Power Apps Portals"'}`]);
+
+        gulp.src('README.md')
+            .pipe(setPreviewState ? replace(standardHeader, previewHeader) : replace(previewHeader, standardHeader))
```
joem-msft, line 229: if you don't take my comment above, at least fix this line. I believe the last call to replace may have incorrect args?

joem-msft: oh, this is a weird side-effect. Essentially, this function is not idempotent; it requires it to have been called previously.
I would consider another way of managing this, perhaps using a backup of the original Readme.md instead in the cleanup logic?

joem-msft (approved):

gulpfile.js
tehcrashxor: the `gulp-replace` logic there is idempotent - it's a find-and-replace, so no replacement occurs if it is already in the correct state

gulpfile.js
tehcrashxor: Updated, but we couldn't be quite as clean as the example you suggested.   
The readme replacement logic needs to not only know what it should place in the file, but also what the other config's string is as a replacement target.

PR #339, opened 2022-11-04
Use node 16.x in official builds, as our Contributing.md already requires

tehcrashxor:
Contributing.md mentions the Node 16.x requirement, but the official build was using 14.
16+ is required for the `npm pkg set` commands in the `packageVSIX` gulp task.

[AB#3010171](https://dev.azure.com/dynamicscrm/1fb98997-2f9e-4734-be8a-9728680447c2/_workitems/edit/3010171)

### DIFF
--- a/.azure-pipelines/OfficialBuild.yml
+++ b/.azure-pipelines/OfficialBuild.yml
@@ -34,9 +34,9 @@ steps:
   displayName: Capture build version in package/version.csv
 
 - task: NodeTool@0
-  displayName: 'Use nodejs 14.x'
+  displayName: 'Use nodejs 16.x'
   inputs:
-    versionSpec: '14.x'
+    versionSpec: '16.x'
 
 - task: Npm@1
   displayName: 'Restore (npm install)'


.azure-pipelines/OfficialBuild.yml
```diff
@@ -34,9 +34,9 @@ steps:
   displayName: Capture build version in package/version.csv
 
 - task: NodeTool@0
-  displayName: 'Use nodejs 14.x'
+  displayName: 'Use nodejs 16.x'
   inputs:
-    versionSpec: '14.x'
+    versionSpec: '16.x'
```
davidjenni, line 39: nodejs 16 (or even 18, the new LTS) is fine as toolchaing to >build< the extension; near term we should change our PR workflow to run our VSCode extension tests on either the min bar version of VSCode (14??), or run tests on a matrix of former and current LTS, and 19?

davidjenni (approved):

.azure-pipelines/OfficialBuild.yml
tehcrashxor: Created a work item to track testing/upgrading to LTS 18

PR #345, opened 2022-11-10
New data model support in web extension

tyaginidhi:


src/web/client/utility/schemaHelper.ts
```diff
@@ -0,0 +1,118 @@
...
+    return { source: attributePathArray[0], relativePath: attributePathArray[1] ?? '' };
+}
+
+export function useBase64Decoding(entity: string, attributeType: string): boolean {
+    return entity === Constants.schemaEntityName.WEBFILES &&
+        (attributeType === Constants.entityAttributesWithBase64Encoding.documentbody || attributeType === Constants.entityAttributesWithBase64Encoding.filecontent);
+}
+
+export function useBase64Encoding(entity: string, attributeType: string): boolean {
+    return entity === Constants.schemaEntityName.WEBFILES &&
```
ameyaapte1, line 38: nit: Function names can be isBase64Decoding or shouldBase64Decode, 'use' doesn't convey it's a boolean return type

src/web/client/utility/UrlBuilder.ts
```diff
@@ -3,34 +3,61 @@
...
             break;
         default:
             break;
     }
 
-    return parameterizedUrlTemplate.replace('{dataverseOrgUrl}', dataverseOrgUrl).replace('{entity}', entity)
-        .replace('{entityId}', entityId).replace('{api}', powerPlatformContext.dataSourcePropertiesMap.get('api') as string)
-        .replace('{data}', powerPlatformContext.dataSourcePropertiesMap.get('data') as string)
-        .replace('{version}', powerPlatformContext.dataSourcePropertiesMap.get('version') as string);
+    return parameterizedUrlTemplate.replace('{dataverseOrgUrl}', dataverseOrgUrl).replace('{entity}', getEntity(entity)?.get(schemaEntityKey.DATAVERSE_ENTITY_NAME) as string)
```
ameyaapte1, line 38: Can we write a formatString utility instead of chaining replace across code?

src/web/client/utility/CommonUtility.ts
```diff
@@ -15,20 +15,16 @@ export function toBase64(data: string) {
     return btoa(unescape(encodeURIComponent(data)));
```
ameyaapte1, line 15: unescape is a deprecated function and should not be used, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape, my siggestion is to use a third party library for base64 encode/decode as btoa breaks for UTF-8 without unescaping, also unescpae ends up omitting some special characters

src/web/client/utility/UrlBuilder.ts
```diff
@@ -3,34 +3,61 @@
...
-        return powerPlatformContext.dataSourcePropertiesMap.get(SINGLE_ENTITY_URL_KEY) as string;
+        return powerPlatformContext.dataSourcePropertiesMap.get(schemaKey.SINGLE_ENTITY_URL) as string;
     }
 
-    return powerPlatformContext.dataSourcePropertiesMap.get(MULTI_ENTITY_URL_KEY) as string;
+    return powerPlatformContext.dataSourcePropertiesMap.get(schemaKey.MULTI_ENTITY_URL) as string;
 };
 
-export function getRequestURL(dataverseOrgUrl: string, entity: string, entityId: string, method: string, isSingleEntity: boolean): string {
+export function getRequestURL(
```
ameyaapte1, line 19: Can we improve on the method name? This sounds like a GET request whereas it seems to generate a URL and return a string, also this file has a patchRequestURL function too

src/web/client/utility/UrlBuilder.ts
```diff
@@ -3,34 +3,61 @@
...
+    const requestUrl = parameterizedUrl.replace('{dataverseOrgUrl}', dataverseOrgUrl)
+        .replace('{entity}', getEntity(entity)?.get(schemaEntityKey.DATAVERSE_ENTITY_NAME) as string)
+        .replace('{api}', powerPlatformContext.dataSourcePropertiesMap.get(schemaKey.API) as string)
+        .replace('{data}', powerPlatformContext.dataSourcePropertiesMap.get(schemaKey.DATA) as string)
+        .replace('{version}', powerPlatformContext.dataSourcePropertiesMap.get(schemaKey.DATAVERSE_API_VERSION) as string);
+
+    return requestUrl + fetchQueryParameters;
+}
+
+export function patchRequestUrl(entity: string, attributeType: string, requestUrl: string) {
```
ameyaapte1, line 57: Can we rename this, it's returning a boolean, maybe isPatchRequest? Also can this be in utility?

src/web/client/common/authenticationProvider.ts
```diff
@@ -7,13 +7,12 @@ import * as vscode from 'vscode';
 import * as nls from 'vscode-nls';
 nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
 import { sendErrorTelemetry, sendInfoTelemetry } from '../telemetry/webExtensionTelemetry';
-import { pathParamToSchema, PROVIDER_ID, telemetryEventNames } from './constants';
-import PowerPlatformExtensionContextManager from "./localStore";
+import { PROVIDER_ID, telemetryEventNames } from './constants';
 
-export function getHeader(accessToken: string) {
+export function getHeader(accessToken: string, useOctetStreamContentType?: boolean) {
```
ameyaapte1, line 12: Should we start using more standard library implementation like axios rather than building headers manually?

src/web/client/common/localStore.ts
```diff
@@ -23,56 +25,62 @@ export interface IPowerPlatformExtensionContext {
     entity: string;
     rootDirectory: vscode.Uri;
     saveDataMap: Map<string, SaveEntityDetails>,
-    defaultFileUri: vscode.Uri // This will default to home page or current page in multifile scenario
+    defaultFileUri: vscode.Uri, // This will default to home page or current page in multifile scenario
+    contextSet: boolean,
+    currentSchema: string
 }
 
 class PowerPlatformExtensionContextManager {
```
ameyaapte1, line 33: Can we match files names and the class in the files, helps understanding/writing code easier?

src/web/client/common/localStore.ts
```diff
@@ -23,56 +25,62 @@ export interface IPowerPlatformExtensionContext {
...
-        this.PowerPlatformExtensionContext = {
-            ...this.PowerPlatformExtensionContext,
-            entity: pathParamToSchema.get(pseudoEntityName) as string,
-            entityId: entityId,
-            queryParamsMap: queryParamsMap,
-            rootDirectory: vscode.Uri.parse(`${PORTALS_URI_SCHEME}:/${queryParamsMap.get(WEBSITE_NAME) as string}/`, true)
-        };
-
-        return this.PowerPlatformExtensionContext;
+    public async setPowerPlatformExtensionContext(entityName: string, entityId: string, queryParamsMap: Map<string, string>) {
```
ameyaapte1, line 57: Rename to setPowerPagesContext? This context is specific to portals and not related with the overall extension

src/web/client/common/localStore.ts
ameyaapte1: Also considering this is a context manager it should be a singleton?

src/web/client/utility/CommonUtility.ts
tyaginidhi: Will take this in a later PR

src/web/client/utility/UrlBuilder.ts
tyaginidhi: This is in utility files.

src/web/client/common/authenticationProvider.ts
tyaginidhi: Out of scope. We can consider this in  later PR.

src/web/client/common/localStore.ts
tyaginidhi: yeah, it is a singleton pattern implemented here. Will take file renames separately. Out of scope for this PR.

src/web/client/common/localStore.ts
tyaginidhi: Not really. This is a extension wide object, and should be used by other apps onboarding to our extension for that purpose.

src/web/client/utility/UrlBuilder.ts
tyaginidhi: We need to revisit naming convention in this extension as a whole. Out of scope for this PR.

src/web/client/utility/UrlBuilder.ts
tyaginidhi: Agree, we need to cleanup this code. Out of scope for this PR.

src/web/client/utility/schemaHelper.ts
tyaginidhi: Agree. Need to revisit this later. Out of scope for current PR.

PR #374, opened 2022-12-05
3076077: Save file telemetry - introduce file extension type

tyaginidhi:
image


amitjoshi438:
Can we just log the file typek? since we are already logging type of entity separately.

tyaginidhi:
> Can we just log the file type? since we are already logging type of entity separately.

this attribute we already maintain as part of portal schema. So we are logging what we are already using. We can sanitize this to extract the last part of the extension like css, html, etc. But lets do this as part of query.

ashishchoudhary001 (approved):

PR #378, opened 2022-12-07
Added new integration_test

v-vbhanndari:


src/web/client/test/integration/urlBuilderUtil.test.ts
```diff
@@ -0,0 +1,337 @@
...
+        } as IPowerPlatformExtensionContext;
+        sinon
+            .stub(
+                PowerPlatformExtensionContextManager,
+                "getPowerPlatformExtensionContext"
+            )
+            .returns(powerPlatformExtensionContext);
+        const isSingleEntity = true;
+        const result = getParameterizedRequestUrlTemplate(isSingleEntity);
+        expect(result).eq("singleEntityURL");
```
tyaginidhi, line 48: why not use constants value for this one like below?

tyaginidhi:
Are all the IT cases? and none UT?

src/web/client/test/integration/urlBuilderUtil.test.ts
v-vbhanndari: Thanks. I will update that in my next PR

tyaginidhi (approved):

PR #390, opened 2022-12-20
added test cases for error handler

v-vbhanndari:
Added new UTs for Error handler


src/web/client/test/integration/errorHandler.test.ts
```diff
@@ -0,0 +1,634 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
```
BidishaMS, line 2: It looks more UT to me rather IT. Please move it under UT folder

src/web/client/test/integration/errorHandler.test.ts
```diff
@@ -0,0 +1,634 @@
...
+    removeEncodingFromParameters,
+    checkMandatoryParameters,
+    checkMandatoryPathParameters,
+    checkMandatoryQueryParameters,
+} from "../../common/errorHandler";
+describe("errorHandler", () => {
+    afterEach(() => {
+        Sinon.restore();
+    });
+    it("showErrorDialog_shouldCallShowErrorMessage", () => {
```
BidishaMS, line 31: I think you didnt follow the naming convention.

image

src/web/client/test/integration/errorHandler.test.ts
```diff
@@ -0,0 +1,634 @@
...
+        const options = { detail: detailMessage, modal: true };
+        assert.calledOnceWithExactly(
+            _mockShowErrorMessage,
+            errorString,
+            options
+        );
+    });
+
+    it("showErrorDialog_shouldCallShowErrorMessage_whenDetailMessagesIsNotPassed", () => {
+        //Act
```
BidishaMS, line 50: WHat happens when errorString is undefined? We should have a test case for that too

src/web/client/test/integration/errorHandler.test.ts
```diff
@@ -0,0 +1,634 @@
...
+            _mockShowErrorMessage,
+            errorString,
+            options
+        );
+    });
+
+    it("removeEncodingFromParameters_shouldDecodeSCHEMAVERSIONAndWEBSITENAME", () => {
+        //Act
+        const queryParamsMap = new Map<string, string>([
+            [schemaKey.SCHEMA_VERSION, "3A%20test_file_name"],
```
BidishaMS, line 69: Why we need 3a? In URL %3a means '/'.
Remove this 3a and keep only test_file_name and test_test_websiteName

src/web/client/test/integration/errorHandler.test.ts
```diff
@@ -0,0 +1,634 @@
...
+        const websiteName = queryParamsMap.get(
+            queryParameters.WEBSITE_NAME
+        ) as string;
+
+        const schemaFileName = queryParamsMap.get(
+            schemaKey.SCHEMA_VERSION
+        ) as string;
+
+        expect(schemaFileName).eq("3A test_file_name");
+        expect(websiteName).eq("3A test_test_webSiteName");
```
BidishaMS, line 86: where is the check for portalFolderName

src/web/client/test/integration/errorHandler.test.ts
```diff
@@ -0,0 +1,634 @@
...
+        const queryParamsMap = new Map<string, string>([
+            [schemaKey.SCHEMA_VERSION, "3A%20test_file_name"],
+            [queryParameters.WEBSITE_NAME, ""],
+        ]);
+
+        //Action
+        removeEncodingFromParameters(queryParamsMap);
+
+        //Assert
+        const websiteName = queryParamsMap.get(
```
BidishaMS, line 100: create another variable with portalfoldername and then do assertions with portalfoldername. Keep websiteName as it is

src/web/client/test/integration/errorHandler.test.ts
```diff
@@ -0,0 +1,634 @@
...
+
+        const schemaFileName = queryParamsMap.get(
+            schemaKey.SCHEMA_VERSION
+        ) as string;
+
+        expect(schemaFileName).eq("3A test_file_name");
+        expect(websiteName).eq(PORTALS_FOLDER_NAME_DEFAULT);
+    });
+
+    it("checkMandatoryParameters_shouldReturnTrueWhenAppNameIsPortalAndEntityAndEntityIdHavingValues", () => {
```
BidishaMS, line 112: checkMandatoryPathParameters

BidishaMS, line 112: AGain Naming convention not followed

src/web/client/test/integration/errorHandler.test.ts
```diff
@@ -0,0 +1,634 @@
...
+        //Act
+        const appName = "portal";
+        const entity = "entity";
+        const entityId = "3";
```
BidishaMS, line 116: entityId's are generally GUID. Keep it similar. 3 does not make any sense.

BidishaMS:
Naming convention is not at all followed. Please correct it in overall PR

src/web/client/test/integration/errorHandler.test.ts
```diff
@@ -0,0 +1,634 @@
...
+
+    it("checkMandatoryPathParameters_shouldReturnFalseWhenEntityIsBlank", () => {
+        //Act
+        const _mockShowErrorDialog = spy(vscode.window, "showErrorMessage");
+        const _mockSendErrorTelemetry = spy(
+            powerPlatformExtensionContext.telemetry,
+            "sendErrorTelemetry"
+        );
+        const appName = "portal";
+        const entity = "";
```
BidishaMS, line 347: What happens when entity is undefined?

src/web/client/test/integration/errorHandler.test.ts
```diff
@@ -0,0 +1,634 @@
...
+        //Action
+        const result = checkMandatoryPathParameters(appName, entity, entityId);
+        //Assert
+        const detailMessage = localize(
+            "microsoft-powerapps-portals.webExtension.init.app-not-found",
+            "Unable to find that app"
+        );
+        expect(result).false;
+        assert.calledWith(_mockShowErrorMessage, detailMessage);
+        assert.calledOnce(_mockShowErrorMessage);
```
BidishaMS, line 420: Why you need to mock error message if appName is not portal. It will only return false in default section

src/web/client/test/integration/errorHandler.test.ts
```diff
@@ -0,0 +1,634 @@
...
+        assert.calledOnceWithExactly(
+            _mockSendErrorTelemetry,
+            telemetryEventNames.WEB_EXTENSION_MANDATORY_PATH_PARAMETERS_MISSING
+        );
+    });
+
+    it("checkMandatoryPathParameters_shouldReturnFalseWhenPppNameIsDifferentFromPortal", () => {
+        //Act
+        const _mockShowErrorMessage = stub(vscode.window, "showErrorMessage");
+        const appName = "por";
```
BidishaMS, line 408: What happens when appname is undefined. I think we should have test case around that also

src/web/client/test/integration/errorHandler.test.ts
```diff
@@ -0,0 +1,634 @@
...
+        const appName = "portal";
+        const queryParamsMap = new Map<string, string>([
+            [queryParameters.ORG_URL, "ORG_URL"],
+            [queryParameters.DATA_SOURCE, ""],
+            [schemaKey.SCHEMA_VERSION, "1.0.0.0"],
+            [
+                queryParameters.WEBSITE_ID,
+                "ed9a6c19-5ab6-4f67-8c35-2423cff958c4",
+            ],
+        ]);
```
BidishaMS, line 501: I see lot of common code when orgURL/ dataSource /schemaName /websiteId is null.
Keep a common method and call it.

src/web/client/test/integration/errorHandler.test.ts
```diff
@@ -0,0 +1,634 @@
...
+        //Assert
+        expect(result).true;
+    });
+
+    it("checkMandatoryParameters_shouldReturnFalseWhenEntityIsBlank", () => {
+        //Act
+        const appName = "portal";
+        const entity = "";
+        const entityId = "3";
+        const queryParamsMap = new Map<string, string>([
```
BidishaMS, line 144: I see lot of common code in each test case. Extract them and write a single method for each common code

src/web/client/test/integration/errorHandler.test.ts
v-vbhanndari: we cannot assign undefined string to errorString

src/web/client/test/integration/errorHandler.test.ts
v-vbhanndari: portalFolderName is the value of "queryParameters.WEBSITE_NAME"

src/web/client/test/integration/errorHandler.test.ts
v-vbhanndari: Done

src/web/client/test/integration/errorHandler.test.ts
v-vbhanndari: Cannot pass string as undefined

src/web/client/test/integration/errorHandler.test.ts
v-vbhanndari: yes but we are asserting the the function calling "_mockShowErrorMessage" without mock we cannot asserts

src/web/client/test/integration/errorHandler.test.ts
v-vbhanndari: Cannot convert string to undefined

src/web/client/test/integration/errorHandler.test.ts
v-vbhanndari: But we are sending different different data in orgURL/ dataSource /schemaName /websiteId

src/web/client/test/integration/errorHandler.test.ts
v-vbhanndari: Different different values are being served

src/web/client/test/integration/errorHandler.test.ts
v-vbhanndari: I am using this because decode URI should remove this 3A from Url and return the decoded url.

v-vbhanndari: If we remove that 3A it will return same string as without 3A

src/web/client/test/integration/errorHandler.test.ts
```diff
@@ -0,0 +1,636 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
```
BidishaMS, line 3: check the file correctly I still see naming convention not followed

src/web/client/test/integration/errorHandler.test.ts
```diff
@@ -0,0 +1,636 @@
...
+} from "../../common/constants";
+
+import {
+    showErrorDialog,
+    removeEncodingFromParameters,
+    checkMandatoryParameters,
+    checkMandatoryPathParameters,
+    checkMandatoryQueryParameters,
+} from "../../common/errorHandler";
+describe("errorHandler", () => {
```
BidishaMS, line 27: newline

BidishaMS (changes requested):

src/web/client/test/integration/errorHandler.test.ts
v-vbhanndari: Sorry it was a old PR so Will take care of it

BidishaMS (approved):

PR #391, opened 2022-12-21
ut for portal schema reader

v-vbhanndari:


src/web/client/test/integration/portalSchemaReader.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    getDataSourcePropertiesMap,
+    getEntitiesFolderNameMap,
+} from "../../../client/schema/portalSchemaReader";
+import { expect } from "chai";
+
+describe("portalSchemaReader", () => {
+    afterEach(() => {
+        Sinon.restore();
+    });
+    it("getEntitiesSchemaMap", () => {
```
BidishaMS, line 21: proper naming of the test case is necessary. Suppose you are testing it whether it is returning map DS for the schema or not then name it accordingly.
getEntitiesSchemaMap_withValidSchema_shouldBeMap

src/web/client/test/integration/portalSchemaReader.test.ts
```diff
@@ -0,0 +1,130 @@
...
+        }
+        const websitesData = result.get("websites");
+        expect(websitesData?.get("_vscodeentityname")).equal("websites");
+        expect(websitesData?.get("_dataverseenityname")).equal(
+            "powerpagesites"
+        );
+        expect(websitesData?.get("_displayname")).equal("Website");
+    });
+
+    it("getEntitiesSchemaMap() - empty entity", () => {
```
BidishaMS, line 39: getEntitiesSchemaMap_withEmptyEntity_shouldBeEmptyMap

src/web/client/test/integration/portalSchemaReader.test.ts
```diff
@@ -0,0 +1,130 @@
...
+        stub(schemaHelperUtil, "getPortalSchema").returns(
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
BidishaMS, line 41: 1) Where is the assert if the schema itself is corrupted?
2) What happens if we send undefined/null schema to the function?
3) What happens if schema length is 0?
4) How you will handle if the return datatype is changed?
5) The schema is so exhaustive why so few metadata is present in your mocked schema?
6) Where you are testing for portal_schema_v1 and portal_Schema_V2
7) Your UT is not sufficient if the schema is changed. Your UT should verify each and every field of the schema

src/web/client/test/integration/portalSchemaReader.test.ts
```diff
@@ -0,0 +1,130 @@
...
+            portalSchemaReadeMock.schema_data_with_empty_array as any
+        );
+
+        const result = getEntitiesSchemaMap("testSchema");
+        for (const [key] of result) {
+            expect(key).equal(undefined);
+        }
+    });
+
+    it("getDataSourcePropertiesMap()", () => {
```
BidishaMS, line 51: I will review other methods only when above comments are addressed.

BidishaMS (changes requested):

src/web/client/test/integration/portalSchemaReader.test.ts
v-vbhanndari: Please Find the Below comments :

1. If the schema is corrupt the function it self throw an error
2. we will not able to send undefined or null because parameter is type of string.
3. Function will throw an exception
4. We don't have any assert to verify the type becuase typeof is also showing Map as 'object'
5. Implemented 
6. That will be test when we will be writing the test case for "schemaHelperUtil"
7. Implemented

src/web/client/test/integration/portalSchemaReader.test.ts
v-vbhanndari: Done

src/web/client/test/integration/portalSchemaReader.test.ts
v-vbhanndari: Done

src/web/client/test/integration/portalSchemaReader.test.ts
```diff
@@ -0,0 +1,225 @@
...
+    getEntitiesFolderNameMap,
+} from "../../../client/schema/portalSchemaReader";
+import { expect, assert } from "chai";
+
+describe("portalSchemaReader", () => {
+    afterEach(() => {
+        Sinon.restore();
+    });
+    it("getEntitiesSchemaMap_withValidSchemaV2_shouldBeMap", () => {
+        //Action
```
BidishaMS, line 22: Don't comment action & assert. Its for internal understanding

src/web/client/test/integration/portalSchemaReader.test.ts
```diff
@@ -0,0 +1,225 @@
...
+        const result = getEntitiesFolderNameMap(entitiesSchemaMap);
+
+        //Assert
+        expect(result?.get("webfiles")).equal("testFolderName");
+        expect(result?.get("webpages")).equal("testFolderName1");
+        expect(result?.get("webtemplates")).equal("testFolderName2");
+    });
+
+    it("getEntitiesFolderNameMap_With_Different_schemaEntityKey_Should_Not_Map", () => {
+        //Act
```
BidishaMS, line 173: getEntitiesFolderNameMap_withDifferentSchemaEntityKey_shouldNotBeMap

src/web/client/test/integration/portalSchemaReader.test.ts
```diff
@@ -0,0 +1,225 @@
...
+
+        //Action
+        const result = getEntitiesFolderNameMap(entitiesSchemaMap);
+
+        //Assert
+        expect(result.size).eq(0);
+    });
+
+    it("getEntitiesFolderNameMap_With_Different_entitiesSchemaKey_Should_Not_Map", () => {
+        //Act
```
BidishaMS, line 200: What is the difference between above and this test case

src/web/client/test/integration/portalSchemaReader.test.ts
v-vbhanndari: ok

src/web/client/test/integration/portalSchemaReader.test.ts
v-vbhanndari: DOne

src/web/client/test/integration/portalSchemaReader.test.ts
v-vbhanndari: In the above one i am sending the "schemaEntityKey.FILE_FOLDER_NAME" as key that will be map in the actual function but in the below test case i am not sending the key as "schemaEntityKey.FILE_FOLDER_NAME" so it wont map the key and will return empty Map

src/web/client/test/integration/portalSchemaReader.test.ts
v-vbhanndari: You want me to have Act , Action and Assert like this Assert , Act , Action . I am doing the same it's showing error.

src/web/client/test/integration/portalSchemaReader.test.ts
BidishaMS: The naming convention is still not correct

v-vbhanndari: Done and changes pushed changes

BidishaMS (approved):

PR #392, opened 2022-12-21
added new test and place test cases in right file

v-vbhanndari:


src/web/client/test/unit/extension.test.ts
```diff
@@ -4,15 +4,9 @@
...
-    it("it should return text from base 64", () => {
-        const str = "this is test case";
-
-        const base64 = convertStringtoBase64(str);
-        const base64totext = convertfromBase64ToString(base64);
-        expect(base64).eq("dGhpcyBpcyB0ZXN0IGNhc2U=");
-        expect(base64totext).eq(str);
+describe("test", () => {
+    it("testing", () => {
+        expect(1).eq(1);
```
BidishaMS, line 10: Is this some placeholder?

src/web/client/test/integration/commonUtil.test.ts
```diff
@@ -0,0 +1,67 @@
...
+import { expect } from "chai";
+import {
+    convertfromBase64ToString,
+    convertStringtoBase64,
+    GetFileNameWithExtension,
+} from "../../utilities/commonUtil";
+
+describe("commonUtil", async () => {
+    it("it should return text from base 64", () => {
+        const str = "this is test case";
```
BidishaMS, line 15: None of the comments are incorporated in this PR. Could you follow the best practices in this PR too which I have shared over mail?

v-vbhanndari: Thanks Bidisha. This PR raised before the comments will implement the comments here as well .

src/web/client/test/unit/extension.test.ts
v-vbhanndari: yes. Initially i removed this file but this file is being used in config so i added and move test cases to it's appropriate location

src/web/client/test/integration/commonUtil.test.ts
```diff
@@ -0,0 +1,105 @@
...
+import { expect } from "chai";
+import { schemaEntityName } from "../../schema/constants";
+import {
+    convertfromBase64ToString,
+    convertStringtoBase64,
+    GetFileNameWithExtension,
+} from "../../utilities/commonUtil";
+
+describe("commonUtil", async () => {
+    it("encodeBase64_And_DecodeBase64_string", () => {
```
BidishaMS, line 15: proper naming missing and don't mix two methods in a single test case.
Write different test cases

v-vbhanndari: ok

BidishaMS:
Please add a snippet of code coverage for your files only.

BidishaMS (approved):

PR #396, opened 2022-12-23
Renamed PowerPlatformExtensionContext to WebExtensionContext

BidishaMS:


src/web/client/WebExtensionContext.ts
```diff
@@ -52,74 +52,74 @@ class WebExtensionContext {
         currentSchemaVersion: ""
     };
 
-    public getPowerPlatformExtensionContext() {
-        return this.powerPlatformExtensionContext;
+    public getWebExtensionContext() {
```
tyaginidhi, line 55: we can actually make this a getter function with name context. In any case we are accessing all of these as WebExtensionContext.<func>. Or, any better suggestions here?

src/web/client/WebExtensionContext.ts
```diff
@@ -52,74 +52,74 @@ class WebExtensionContext {
...
+        return this.webExtensionContext;
     }
 
-    public async setPowerPlatformExtensionContext(entityName: string, entityId: string, queryParamsMap: Map<string, string>) {
+    public setWebExtensionContext(entityName: string, entityId: string, queryParamsMap: Map<string, string>) {
```
tyaginidhi, line 59: this can be a setter function too

tyaginidhi (approved):

src/web/client/WebExtensionContext.ts
BidishaMS: As discussed created a task https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/3116420

src/web/client/WebExtensionContext.ts
BidishaMS: https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/3116420

PR #400, opened 2023-01-02
Added nyc lib which is a successor of istanbul for Code coverage

BidishaMS:
To run coverage use the command 'npm run coverage'

image
image


tyaginidhi (approved):

tyaginidhi:
we should consider using "gulp testCoverage" command as well here

tyaginidhi (approved):

PR #418, opened 2023-01-17
webpage create integration

amitjoshi438:


src/client/powerpages/utils/CommonUtils.ts
```diff
@@ -17,3 +21,88 @@ export const fileName = (name: string) => {
     // Uppercase the first letter of each word and join the words back together
     return words.map((word) => word[0].toUpperCase() + word.slice(1)).join("-");
 };
+
+interface pageTemplate {
+    name: string;
+    value: string;
+}
```
BidishaMS, line 28: Why interface in commonutil?

src/client/powerpages/utils/CommonUtils.ts
```diff
@@ -17,3 +21,88 @@ export const fileName = (name: string) => {
...
+            }
+            prevPage = page;
+            page = pages.get(page.adx_parentpageid);
+            path = `${page.adx_name}/${path}`;
+        }
+        if (paths.indexOf(path) === -1) {
+            paths.push(path);
+            pathsMap.set(path, webpageid);
+        }
+    }
```
BidishaMS, line 105: Iterative loop is not so much recommended. We should use recursion.

PR #419, opened 2023-01-17
Origin/users/amitjoshi/content snippet

amitjoshi438:
This PR integrates content snippet creation in VSCode using Yeoman Generator:
contensnippetCreation

### DIFF
--- a/src/client/powerpages/Contentsnippet.ts
+++ b/src/client/powerpages/Contentsnippet.ts
@@ -10,19 +10,19 @@ nls.config({
 })();
 const localize: nls.LocalizeFunc = nls.loadMessageBundle();
 import * as vscode from "vscode";
-import { fileName, folderName, isNullOrEmpty } from "./utils/CommonUtils";
+import { formatFileName, formatFolderName, isNullOrEmpty } from "./utils/CommonUtils";
 import { QuickPickItem } from "vscode";
 import { MultiStepInput } from "./utils/MultiStepInput";
 
-export const contentSnippet = async (context: vscode.ExtensionContext) => {
-    const { name, type } = await myMultiStepInput();
+export const createContentSnippet = async (context: vscode.ExtensionContext) => {
+    const { contentSnippetName, contentSnippetType } = await getContentSnippetInputs();
 
-    if (!isNullOrEmpty(name)) {
+    if (!isNullOrEmpty(contentSnippetName)) {
         const terminal = vscode.window.createTerminal("Powerpages", "");
-        terminal.sendText(`cd data\n ../node_modules/.bin/yo @microsoft/powerpages:contentsnippet "${name}" "${type}"`);
+        terminal.sendText(`cd data\n ../node_modules/.bin/yo @microsoft/powerpages:contentsnippet "${contentSnippetName}" "${contentSnippetType}"`);
 
-        const folder = folderName(name);
-        const file = fileName(name);
+        const folder = formatFolderName(contentSnippetName);
+        const file = formatFileName(contentSnippetName);
 
         const watcher = vscode.workspace.createFileSystemWatcher(
             new vscode.RelativePattern(
@@ -51,7 +51,7 @@ export const contentSnippet = async (context: vscode.ExtensionContext) => {
 };
 
 
-async function myMultiStepInput() {
+async function getContentSnippetInputs() {
     const contentSnippetTypes: QuickPickItem[] = ["html", "text"].map(
         (label) => ({ label })
     );
@@ -60,8 +60,8 @@ async function myMultiStepInput() {
         title: string;
         step: number;
         totalSteps: number;
-        type: QuickPickItem | string;
-        name: string;
+        contentSnippetType: QuickPickItem | string;
+        contentSnippetName: string;
     }
 
     async function collectInputs() {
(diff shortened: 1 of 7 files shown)


src/client/powerpages/Contentsnippet.ts
```diff
@@ -0,0 +1,118 @@
...
+        terminal.sendText(`cd data\n ../node_modules/.bin/yo @microsoft/powerpages:contentsnippet "${name}" "${type}"`);
+
+        const folder = folderName(name);
+        const file = fileName(name);
+
+        const watcher = vscode.workspace.createFileSystemWatcher(
+            new vscode.RelativePattern(
+                // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                vscode.workspace.workspaceFolders![0],
+                `**/content-snippets/${folder}/${file}.en-US.contentsnippet.yml`
```
ashishchoudhary001, line 31: the language value here en-US shouldn't be hardcoded.

src/client/powerpages/Contentsnippet.ts
```diff
@@ -0,0 +1,118 @@
...
+import * as vscode from "vscode";
+import { fileName, folderName, isNullOrEmpty } from "./utils/CommonUtils";
+import { QuickPickItem } from "vscode";
+import { MultiStepInput } from "./utils/MultiStepInput";
+
+export const contentSnippet = async (context: vscode.ExtensionContext) => {
+    const { name, type } = await myMultiStepInput();
+
+    if (!isNullOrEmpty(name)) {
+        const terminal = vscode.window.createTerminal("Powerpages", "");
```
BidishaMS, line 21: Should'nt this be localized?

src/client/powerpages/Contentsnippet.ts
```diff
@@ -0,0 +1,118 @@
...
+            watcher.onDidCreate(async (uri) => {
+                await vscode.window.showTextDocument(uri);
+                terminal.dispose();
+            });
+        });
+    }
+};
+
+
+async function myMultiStepInput() {
```
BidishaMS, line 54: Please rename the method

src/client/powerpages/Contentsnippet.ts
```diff
@@ -0,0 +1,118 @@
...
+        (label) => ({ label })
+    );
+
+    interface State {
+        title: string;
+        step: number;
+        totalSteps: number;
+        type: QuickPickItem | string;
+        name: string;
+    }
```
BidishaMS, line 65: Looks like generic interface. Please make a separate file for this

src/client/powerpages/utils/MultiStepInput.ts
```diff
@@ -0,0 +1,190 @@
...
+                this.current = input;
+                this.current.show();
+            });
+        } finally {
+            disposables.forEach((d) => d.dispose());
+        }
+    }
+
+    async showInputBox<P extends InputBoxParameters>({
+        title,
```
BidishaMS, line 126: Naming convention looks very naive in most of the method

src/client/powerpages/Contentsnippet.ts
```diff
@@ -0,0 +1,118 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
```
BidishaMS, line 2: ContentSnippet does not looks descriptive. How we name it as ContentSnippetOperations

src/client/powerpages/utils/CommonUtils.ts
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ */
+
+export const isNullOrEmpty = (str: string | undefined): boolean => {
```
BidishaMS, line 6: There is a significant difference in null and undefined in JS. 
Keep a method for the check null and undefined.
keep a method for empty length string.

Incase you are interested to understand the difference https://www.geeksforgeeks.org/undefined-vs-null-in-javascript/

src/client/powerpages/utils/CommonUtils.ts
```diff
@@ -0,0 +1,19 @@
...
+    return !str || str.trim().length === 0;
+};
+
+export const folderName = (name: string) => {
+    return name.replace(/[/ ]/g, "-").toLowerCase();
+};
+
+export const fileName = (name: string) => {
```
BidishaMS, line 14: Can you write a comment which shows how the input/output looks like

src/client/powerpages/Contentsnippet.ts
amitjoshi438: I thought proper nouns are not localized, do we localize them also?

src/client/powerpages/Contentsnippet.ts
amitjoshi438: some properties are generic but as such it is specific to entities

src/client/powerpages/utils/MultiStepInput.ts
amitjoshi438: I have updated some custom method names but for vscode related methods I tried sticking to their docs and example implementations. Let me know if some specific method name should be changed.

src/client/powerpages/utils/CommonUtils.ts
amitjoshi438: right, but here '!str' checks for all kind of **falsy** values : false, 0, "", null, undefined, NaN. Let me know if I am missing something.

PR #430, opened 2023-01-25
[rel/stable] Fix web ext telemetry imports

davidjenni:
Cherry pick of #429


davidjenni:
tentatively merging to rel/stable to be able to produce a new candidate build
CC @tyaginidhi

PR #436, opened 2023-02-01
Users/v vbhanndari/remote save provider

v-vbhanndari:


package.json
```diff
@@ -770,7 +770,7 @@
     "gulp-typescript": "^6.0.0-alpha.1",
     "mocha": "^9.2.2",
     "moment": "^2.29.4",
-    "node-fetch": "^2.6.7",
+    "node-fetch": "^2.6.8",
```
BidishaMS, line 773: What's the need of increasing the version?

package.json
```diff
@@ -789,6 +789,7 @@
     "yargs": "^16.2.0"
   },
   "dependencies": {
+    "@types/node-fetch": "^2.6.2",
```
BidishaMS, line 792: Why the version different from devdependencies

package.json
v-vbhanndari: I just ran npm install node-fetch and it took latest version

package.json
v-vbhanndari: That is Type of so i think took latest by npm install

PR #439, opened 2023-02-06
Rename action support for power pages entities

tyaginidhi:
RenameFilesDesktopExtesnsion

Update: Fixed the issues in page-template, and polls entity adx_name updates

TODO: Need to add telemetry and localization updates
Verify yml file field updates for each entity

### DIFF
--- a/src/client/power-pages/commonUtility.ts
+++ b/src/client/power-pages/commonUtility.ts
@@ -4,7 +4,7 @@
  */
 
 import * as vscode from "vscode";
-import { EntityFolderMap, EntityFolderName, PowerPagesEntityType, WebFilesFolder, WebFileYmlExtension } from "./constants";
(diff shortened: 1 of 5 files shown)


amitjoshi438:
We may need to check once; our name formatting is in parity with PAC otherwise on download they will be overwritten.

amitjoshi438:
We may need to check once; our name formatting is in parity with PAC otherwise on download they will be overwritten.

amitjoshi438:
In case of webfile the key **filename** remains same?

amitjoshi438 (approved):

ashishchoudhary001 (approved):

tyaginidhi:
> 

Yeah. that is correct. For webfile the adx_name param has the complete filename like ABC.png, XYZ.mp4 etc. For other entities it varies, and thats where we need to update adx_name update function. I have noticed / and . and - and spaces being set in adx_name.

PR #450, opened 2023-02-17
3079127 - Load NPS Form

BidishaMS:
As of now I am adding screenshot because wiring with current editor will be taken care in upcoming PR

image


tyaginidhi:
Since this is a UX change - can we add a gif for the rendered form in description. Also, a good practice to have ADO work item # in the PR here for easy tracking. Same with linking ADO work item to the PR here.

tyaginidhi:
Also, we should not have webview file in the main web folder - this should ideally go in a new folder called something like webview - so that we can keep on adding new views to this folder in future.

tyaginidhi (approved):

PR #460, opened 2023-02-23
3152262 Change failed to find app error from notification to modal

tyaginidhi:
Show modal instead of notification in case of app load fail scenarios

image


amitjoshi438 (approved):

src/web/client/extension.ts
```diff
@@ -123,6 +120,16 @@ export function processWillSaveDocument(context: vscode.ExtensionContext) {
     );
 }
 
+export async function showSiteVisibilityDialog() {
+    if (WebExtensionContext.urlParametersMap.get(queryParameters.SITE_VISIBILITY) === PUBLIC) {
+        const edit: vscode.MessageItem = { isCloseAffordance: true, title: localize("microsoft-powerapps-portals.webExtension.init.sitevisibility.edit", "Edit the site") };
+        const siteMessage = localize("microsoft-powerapps-portals.webExtension.init.sitevisibility.edit.desc",
+            "Be careful making changes. Anyone can see the changes you make immediately. Choose Edit the site to make edits, or close the editor tab to cancel without editing.");
```
BidishaMS, line 127: should'nt we scrub all these strings?

BidishaMS (approved):

src/web/client/extension.ts
tyaginidhi: Already scrubbed - didnt add anything new here.

PR #466, opened 2023-02-27
added test cases for FilrSystemUpdatesUtility

v-vbhanndari:


package.json
```diff
@@ -778,6 +779,7 @@
     "os-browserify": "^0.3.0",
     "path-browserify": "^1.0.1",
     "process": "^0.11.10",
+    "proxyquire": "^2.1.3",
```
BidishaMS, line 782: why we need this in prod dependencies?

BidishaMS (changes requested):

PR #468, opened 2023-03-01
3158869, 3158870: Localization, telemetry and case-insensitive pattern check enhancement

tyaginidhi:


src/client/extension.ts
```diff
@@ -115,7 +115,7 @@ export async function activate(
     const areCRUDoperationEnabled = vscode.workspace.getConfiguration(SETTINGS_EXPERIMENTAL_STORE_NAME).get(PORTAL_CRUD_OPERATION_SETTING_NAME);
     if (areCRUDoperationEnabled) {
```
amitjoshi438, line 116: can we remove this FF now?

amitjoshi438 (approved):

src/client/extension.ts
tyaginidhi: sure - let me take that up in a separate PR

PR #475, opened 2023-03-10
Powerpages Create Webtemplate Integration

amitjoshi438:
- Integrated Webtemplate command
- Telemetry and localization for it.

### DIFF
--- a/src/client/power-pages/create/CreateCommandWrapper.ts
+++ b/src/client/power-pages/create/CreateCommandWrapper.ts
@@ -5,39 +5,46 @@
 
 import * as vscode from "vscode";
 import { GeneratorAcquisition } from "../../lib/GeneratorAcquisition";
+import { ITelemetry } from "../../telemetry/ITelemetry";
+import { sendTelemetryEventWithTriggerPoint } from "../telemetry";
 import { createContentSnippet } from "./Contentsnippet";
-import { getSelectedWorkspaceFolder } from "./utils/CommonUtils";
+import { Tables} from "./CreateOperationConstants";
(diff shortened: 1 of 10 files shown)


src/client/power-pages/telemetry.ts
```diff
@@ -59,8 +63,8 @@ export function sendTelemetryEvent(telemetry: ITelemetry, telemetryData: IPowerP
 
 export function sendTelemetryEventWithTriggerPoint(entityType: Tables, uri:vscode.Uri, telemetry:ITelemetry): void {
```
tyaginidhi, line 64: @amitjoshi438 Do we really need this extra function in telemetry? Why not use a helper function like uri ? TriggerPoint1: TriggerPoint2 and send that directly in the telemetrydataProperties using the original telemetry function?

tyaginidhi (changes requested):

src/client/power-pages/telemetry.ts
amitjoshi438: right, we can avoid that extra function.

src/client/power-pages/create/CreateCommandWrapper.ts
```diff
@@ -48,4 +50,22 @@ function registerCreateCommands(
             );
         }
     );
+
+    vscode.commands.registerCommand(
+        "microsoft-powerapps-portals.webtemplate",
+        async (uri) => {
+            const triggerPoint = uri?TriggerPoint.CONTEXT_MENU:TriggerPoint.COMMAND_PALETTE;
```
tyaginidhi, line 57: add spacing for readbility

PR #525, opened 2023-05-09
diff view - handle etag update after save

tyaginidhi:
This addresses corner scenario where difference in etag of a recently saved entity triggers un-necessary diff view

### DIFF
--- a/src/web/client/dal/fileSystemProvider.ts
+++ b/src/web/client/dal/fileSystemProvider.ts
@@ -24,6 +24,7 @@ import {
     getEntityEtag,
     getFileEntityEtag,
     getFileEntityId,
+    getFileEntityType,
     updateEntityEtag,
     updateFileDirtyChanges,
 } from "../utilities/fileAndEntityUtil";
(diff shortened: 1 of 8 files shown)


src/web/client/dal/fileSystemProvider.ts
```diff
@@ -74,19 +75,27 @@ export class PortalsFS implements vscode.FileSystemProvider {
...
-            .getConfiguration(SETTINGS_EXPERIMENTAL_STORE_NAME)
-            .get(VERSION_CONTROL_FOR_WEB_EXTENSION_SETTING_NAME);
+        if (fileHasDirtyChanges(uri.fsPath)) {
+            WebExtensionContext.telemetry.sendInfoTelemetry(
+                telemetryEventNames.WEB_EXTENSION_FILE_HAS_DIRTY_CHANGES
+            );
+
+            const isVersionControlEnabled = vscode.workspace
+                .getConfiguration(SETTINGS_EXPERIMENTAL_STORE_NAME)
+                .get(VERSION_CONTROL_FOR_WEB_EXTENSION_SETTING_NAME);
```
BidishaMS, line 85: I think you should a utility method to checkVersionControlEnabled.

src/web/client/dal/fileSystemProvider.ts
```diff
@@ -74,19 +75,27 @@ export class PortalsFS implements vscode.FileSystemProvider {
...
+
+            if (!isVersionControlEnabled) {
+                WebExtensionContext.telemetry.sendInfoTelemetry(
+                    telemetryEventNames.WEB_EXTENSION_DIFF_VIEW_FEATURE_FLAG_DISABLED
```
BidishaMS, line 89: Why we even want to log this? This looks redundant when we make it enabled.
Is it inferring to something which I am missing

tyaginidhi: ideally we would want to know that there are users who explicitly disabled this feature - specially since we are rolling this behind a feature flag to allow the flexibility to disable

BidishaMS (approved):

PR #530, opened 2023-05-16
Introduce continuation token to fetch files for large sites

tyaginidhi:
Dataverse odata calls response with "@odata.nextLink" gives the option to fetch all associated entities for a large portal. Introducing this change here behind multi-file feature flag.

### DIFF
--- a/src/web/client/common/authenticationProvider.ts
+++ b/src/web/client/common/authenticationProvider.ts
@@ -3,63 +3,121 @@
  * Licensed under the MIT License. See License.txt in the project root for license information.
  */
 
-import * as vscode from 'vscode';
-import WebExtensionContext from '../WebExtensionContext';
-import { telemetryEventNames } from '../telemetry/constants';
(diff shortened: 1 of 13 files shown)


src/web/client/common/authenticationProvider.ts
```diff
@@ -3,63 +3,121 @@
  * Licensed under the MIT License. See License.txt in the project root for license information.
  */
```
BidishaMS, line 4: I see you have touched lot of core files and I am pretty sure lot of UTs are affected. What's your plan on adding UTs for them?

BidishaMS: Can we make a practice to add UTs for all newly added methods other it will always in a catch-up game for writing UTs.

BidishaMS (changes requested):

src/web/client/common/authenticationProvider.ts
tyaginidhi: Will take this up as follow up task

BidishaMS:
Approving this PR on the basis that we will get follow-up UTs PR asap. @ashishchoudhary001

BidishaMS (approved):

PR #563, opened 2023-06-22
Support loading all webfiles - old and new data model

tyaginidhi:
Webfile support all files in old and new data model
ODM_SupportMediaFileInExplorer
NDM_SupportMediaFileInExplorer

File content can be lazy loaded for new data model as that is an explicit call to get content. For media files this content is marked for on-click load as this doesn't impact time-to-search.

NOTE: There is an issue in loading content of media files for preview that will be addressed in a separate PR.

### DIFF
--- a/src/web/client/dal/fileSystemProvider.ts
+++ b/src/web/client/dal/fileSystemProvider.ts
@@ -205,7 +205,7 @@ export class PortalsFS implements vscode.FileSystemProvider {
                 }
             );
         }
-
+        
         entry.mtime = Date.now();
         entry.size = content.byteLength;
         entry.data = content;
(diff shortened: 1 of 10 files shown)


src/web/client/dal/fileSystemProvider.ts
```diff
@@ -205,7 +205,7 @@ export class PortalsFS implements vscode.FileSystemProvider {
                 }
             );
         }
-
+
```
BidishaMS, line 208: undo this.

src/web/client/utilities/commonUtil.ts
```diff
@@ -114,4 +117,28 @@ export function isVersionControlEnabled() {
 /* eslint-disable @typescript-eslint/no-explicit-any */
 export function isNullOrUndefined(object: any | null | undefined): boolean {
     return object === null || object === undefined;
+}
+
+// Clean up the file name to remove special characters
+export function getSanitizedFileName(fileName: string): string {
```
BidishaMS, line 123: Can you put a comment what is input and output of this API

BidishaMS (approved):

PR #564, opened 2023-06-23
EtagHandlerService Improvements Exception Handling

BidishaMS:
EtagHandlerService Improvements Exception Handling

1) Catch block has seggregation for API and System failures 
2) Started logging statusCode 
3) Logging eventName previously it was only APIFailure eventName


src/web/client/services/etagHandlerService.ts
```diff
@@ -114,14 +125,25 @@ export class EtagHandlerService {
...
         } catch (error) {
-            const authError = (error as Error)?.message;
-            WebExtensionContext.telemetry.sendAPIFailureTelemetry(
-                requestUrl,
-                entityName,
-                httpMethod.GET,
-                new Date().getTime() - requestSentAtTime,
-                authError
-            );
+            if ((error as Response)?.status>0){
```
tyaginidhi, line 128: nit: formatting

tyaginidhi (approved):

src/web/client/services/etagHandlerService.ts
BidishaMS: doing in next PR

PR #577, opened 2023-07-03
Added missing URL params in telemetry logs

BidishaMS:
Added missing URL params in telemetry logs


src/web/client/common/errorHandler.ts
```diff
@@ -143,7 +144,8 @@ export function checkMandatoryQueryParameters(
                 return true;
             } else {
                 WebExtensionContext.telemetry.sendErrorTelemetry(
-                    telemetryEventNames.WEB_EXTENSION_MANDATORY_QUERY_PARAMETERS_MISSING
+                    telemetryEventNames.WEB_EXTENSION_MANDATORY_QUERY_PARAMETERS_MISSING,
+                    `orgURL:${orgURL}, dataSource:${dataSource}, schemaName:${schemaName} ,websiteId:${websiteId}`
```
tyaginidhi, line 148: Confirm this orgUrl logging for compliance before checking in the code.

tyaginidhi (approved):

src/web/client/common/errorHandler.ts
BidishaMS: We do log orgURL in ppux so I believe it should be compliant. Though I will check it with Dileep once.

tyaginidhi: Correct - our telemetry DB and DAX DB could have difference here.

PR #580, opened 2023-07-04
Support media file preview and fix no content error on empty payload return

tyaginidhi:
image

### DIFF
--- a/src/web/client/utilities/commonUtil.ts
+++ b/src/web/client/utilities/commonUtil.ts
@@ -16,16 +16,23 @@ import { schemaEntityName } from "../schema/constants";
 import { telemetryEventNames } from "../telemetry/constants";
 import WebExtensionContext from "../WebExtensionContext";
 import { SETTINGS_EXPERIMENTAL_STORE_NAME } from "../../../client/constants";
-import { decode, encode } from 'js-base64';
 
-// decodes base64 to text
-export function convertfromBase64ToString(data: string) {
-  return decode(data);
+// decodes file content to UTF-8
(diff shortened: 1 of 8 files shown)


amitjoshi438 (approved):

src/web/client/utilities/commonUtil.ts
```diff
@@ -16,16 +16,23 @@ import { schemaEntityName } from "../schema/constants";
...
-// decodes base64 to text
-export function convertfromBase64ToString(data: string) {
-  return decode(data);
+// decodes file content to UTF-8
+export function convertContentToUint8Array(content: string, isBase64Encoded: boolean): Uint8Array {
+    if(isBase64Encoded){
+        return new Uint8Array(Buffer.from(content, BASE_64));
+    }
+
+    return new TextEncoder().encode(content as string);
```
BidishaMS, line 26: It can be single line
return isBase64Encoded? new Uint8Array(Buffer.from(content, BASE_64)): new TextEncoder().encode(content as string);

src/web/client/utilities/commonUtil.ts
```diff
@@ -16,16 +16,23 @@ import { schemaEntityName } from "../schema/constants";
...
 }
 
-// encodes text to UTF-8 bytes which are then encoded to base64
-export function convertStringtoBase64(data: string) {
-  return encode(data);
+// encodes file content to base64 or returns the content as is
+export function convertContentToString(content: string, isBase64Encoded: boolean): string {
+    if(isBase64Encoded) {
```
BidishaMS, line 31: same here

PR #592, opened 2023-07-10
Remove WEB_EXTENSION_DATAVERSE_SAVE_FILE_FAILED from remoteSaveProvider and remoteFetchProvider

BidishaMS:
This WEB_EXTENSION_DATAVERSE_SAVE_FILE_FAILED is getting logged on success also. Removing it to reduce the wrong logging


src/web/client/dal/remoteSaveProvider.ts
```diff
@@ -189,9 +187,6 @@ async function saveDataToDataverse(
                 new Date().getTime() - requestSentAtTime,
                 fileExtensionType
             );
```
tyaginidhi, line 189: We need to replace this with a success event right?

src/web/client/dal/remoteSaveProvider.ts
```diff
@@ -166,6 +166,7 @@ async function saveDataToDataverse(
...
 
             if (!response.ok) {
+                // TODO: Remove it when multifile stabilizes.
```
tyaginidhi, line 169: remove this one - just verified that all that we are capturing here will be captured in final catch in any case

tyaginidhi (changes requested):

src/web/client/dal/remoteSaveProvider.ts
BidishaMS: I will add the event name in successTelemetry itself rather than creating another event.

tyaginidhi (approved):

PR #594, opened 2023-07-10
Enhancing Error events for GET and UPDATE call to dataverse

BidishaMS:


src/web/client/WebExtensionContext.ts
```diff
@@ -387,8 +387,9 @@ class WebExtensionContext implements IWebExtensionContext {
                     new Date().getTime() - requestSentAtTime,
                     response?.statusText,
                     '',
+                    telemetryEventNames.WEB_EXTENSION_API_REQUEST_FAILURE,
+                    response?.status.toString(),
                     telemetryEventNames.WEB_EXTENSION_POPULATE_LANGUAGE_ID_TO_CODE_API_ERROR,
```
tyaginidhi, line 392: we should not be using the telemetry event name as method name.

src/web/client/dal/remoteFetchProvider.ts
```diff
@@ -147,8 +147,9 @@ async function fetchFromDataverseAndCreateFiles(
                     new Date().getTime() - requestSentAtTime,
                     errorMsg,
                     '',
-                    telemetryEventNames.WEB_EXTENSION_FETCH_DATAVERSE_AND_CREATE_FILES_API_ERROR,
-                    (error as Response)?.status.toString()
+                    telemetryEventNames.WEB_EXTENSION_API_REQUEST_FAILURE,
+                    (error as Response)?.status.toString(),
+                    telemetryEventNames.WEB_EXTENSION_FETCH_DATAVERSE_AND_CREATE_FILES_API_ERROR
```
tyaginidhi, line 152: same here and in all the updates above

src/web/client/telemetry/webExtensionTelemetry.ts
```diff
@@ -86,7 +86,9 @@ export class WebExtensionTelemetry {
         duration?: number,
         errorMessage?: string,
         eventName?: string,
-        status?: string) {
+        status?: string,
+        methodName?: string) {
```
tyaginidhi, line 90: this function signature update is correct - but we should make sure we are passing an actual methodname here - function.name property in javascript

tyaginidhi (changes requested):

src/web/client/WebExtensionContext.ts
BidishaMS: fixed

tyaginidhi (approved):

PR #606, opened 2023-07-12
[PowerPages Copilot] Feedback Form Handling

amitjoshi438:
- Now only a single instance of feedback form will be visible.
- Thumbs up/down will also go as part of feedback


package.json
```diff
@@ -674,7 +670,7 @@
         },
         {
           "command": "powerpages.copilot.clearConversation",
-          "when": "!virtualWorkspace"
+          "when": "never"
```
ashishchoudhary001, line 673: What is this change? Are we removing the option to clear conversation?

amitjoshi438: No, just removing it from command pallete.

ashishchoudhary001 (approved):

PR #609, opened 2023-07-13
[Critical Fix] Include return statement if error telemetry is logged

BidishaMS:
Found out there are few events where we are logging both error and success telemetry. During analysis it is giving a wrong impression. I have added a return statement whenever error telemetry is logged.


src/web/client/dal/remoteSaveProvider.ts
```diff
@@ -167,7 +167,17 @@ async function saveDataToDataverse(
             );
 
             if (!response.ok) {
-                throw new Error(response.statusText);
+                WebExtensionContext.telemetry.sendAPIFailureTelemetry(
```
tyaginidhi, line 170: why are we updating this catch extra telemetry for API here - just throw the entire response object and the catch block will take care of rest.

tyaginidhi: In fact simple return will confuse the FS to believe that the save to dataverse was successful and this will end up getting counted as success scenario - we don't want that!

src/web/client/WebExtensionContext.ts
```diff
@@ -391,6 +391,7 @@ class WebExtensionContext implements IWebExtensionContext {
                     '',
                     response?.status.toString()
                 );
+                return;
```
tyaginidhi, line 394: infact same for these return - just throw the response as error and catch will do the rest! We can remove this extra API logging from these 3 functions

tyaginidhi (changes requested):

tyaginidhi (approved):

PR #686, opened 2023-09-06
Added OrgURLs in customDimension to have better Vscode Desktop Usage count and Copilot Telemetry Improvements

BidishaMS:

PR #689, opened 2023-09-07
Users/biddas/telemetry improvements

BidishaMS:
Cherry-picked Telemetry improvements from master

https://github.com/microsoft/powerplatform-vscode/pull/686

### DIFF
--- a/src/client/extension.ts
+++ b/src/client/extension.ts
@@ -176,6 +176,7 @@ export async function activate(
         _telemetry.sendTelemetryEvent("PowerPagesWebsiteYmlExists"); // Capture's PowerPages Users
         vscode.commands.executeCommand('setContext', 'powerpages.websiteYmlExists', true);
         initializeGenerator(_context, cliContext, _telemetry); // Showing the create command only if website.yml exists
+        
         showNotificationForCopilot(_telemetry, telemetryData, listOfActivePortals.length.toString());
     }
     else {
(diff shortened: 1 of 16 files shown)


src/client/extension.ts
```diff
@@ -21,7 +21,7 @@ import {
     TransportKind,
     WorkspaceFolder,
 } from "vscode-languageclient/node";
-import { workspaceContainsPortalConfigFolder } from "../common/PortalConfigFinder";
+import { getPortalsOrgURLs, workspaceContainsPortalConfigFolder } from "../common/PortalConfigFinder";
```
petrochuk, line 24: How is telemetry related to QFE?

petrochuk (changes requested):

src/client/extension.ts
```diff
@@ -341,16 +352,19 @@ function handleWorkspaceFolderChange() {
     }
 }
 
-function showNotificationForCopilot(telemetry: TelemetryReporter) {
+function showNotificationForCopilot(telemetry: TelemetryReporter, telemetryData:string, countOfActivePortals: string) {
```
joem-msft, line 355: nit: rename `telemetryData` to `listOfOrgsJson` so it's clearer what is expected.

joem-msft, line 355: nit: consider changing type of `countOfActivePorals` to be an integer or number to keep strong typing. Then convert to string only at the site needed.

PR #692, opened 2023-09-08
Introducing Co-Presence on vscode web - infra PR

ritikramuka:
### Introducing Co-Presence on vscode web

# Items done in this PR
- Added feature flag for enabling co-presence in vscode web
- Created treeWebView for showing users across studio and vscode web in the explorer panel
- Added web worker for running co-presence worker script and wired the audience present in the container with treeWebView

image
image
image


tyaginidhi:
Add description about what changes we are pushing as part of this PR and update the heading as well.

package.json
```diff
@@ -433,6 +433,11 @@
           "type": "boolean",
           "markdownDescription": "Enable multiple file view in Visual Studio Code (Web extension only)",
           "default": true
+        },
+        "powerPlatform.experimental.enableCoPresenceFeature": {
+          "type": "boolean",
+          "markdownDescription": "Enable co-presence in Visual Studio Code (Web extension only)",
```
tyaginidhi, line 439: Review this once with PM for text

src/web/client/WebExtensionContext.ts
```diff
@@ -146,6 +157,21 @@ class WebExtensionContext implements IWebExtensionContext {
     public get showMultifileInVSCode() {
         return this._showMultifileInVSCode;
     }
+    public get treeWebView() {
```
tyaginidhi, line 160: nit: keep all the gets together and sets together per previous formatting

src/web/client/WebExtensionContext.ts
```diff
@@ -56,6 +58,8 @@ export interface IWebExtensionContext {
     defaultFileUri: vscode.Uri; // This will default to home page or current page in multifile scenario
     showMultifileInVSCode: boolean;
     extensionActivationTime: number;
+    treeWebView: TreeWebViewProvider;
```
tyaginidhi, line 61: Do we really need this to be added in context interface?

src/web/client/context/userDataMap.ts
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ */
+
+export interface IUserData {
+    containerId: string;
+    fileName: string;
```
tyaginidhi, line 8: Do fileName and filepath need to be array? We will support user support across multiple files right - how are we handling surfacing that information?

src/web/client/context/userDataMap.ts
```diff
@@ -0,0 +1,81 @@
...
+    ) {
+        const userData = new UserData(
+            containerId,
+            fileName,
+            filePath,
+            userName,
+            userId
+        );
+
+        this.usersMap.set(userId, userData);
```
tyaginidhi, line 75: If we are treating userId as key  -we should consider splitting above interface and keeping an array in Map value. This can be done in follow-up PR

src/web/client/extension.ts
```diff
@@ -254,6 +258,125 @@ export function processWillSaveDocument(context: vscode.ExtensionContext) {
     );
 }
 
+let copresenceWorker: Worker;
```
tyaginidhi, line 261: this needs to shift as retrun parameter of function processOpenActiveTextEditor(), correct?

src/web/client/extension.ts
```diff
@@ -254,6 +258,125 @@ export function processWillSaveDocument(context: vscode.ExtensionContext) {
...
+                                filePath: activeEditor.document.uri.fsPath,
+                            },
+                            containerId: WebExtensionContext.containerId,
+                        } as IContainerData);
+                    }
+                }
+            })
+        );
+    } catch (error) {
+        console.log(error);
```
tyaginidhi, line 376: remove the console log from here - this should go to error telemetry function

src/web/client/services/copresenceWorker.js
```diff
@@ -0,0 +1,161 @@
...
+        });
+    } catch (error) {
+        console.log(`Error retrieving container: ${error}`);
+    }
+}
+
+function runFluidApp() {
+    // Listen for messages from the extension
+    // eslint-disable-next-line no-undef
+    self.addEventListener("message", async (event) => {
```
tyaginidhi, line 149: In follow-up PR we should revisit what all telemetry we need to log form this script.

src/web/client/utilities/urlBuilderUtil.ts
```diff
@@ -227,4 +227,42 @@ export function isWebFileWithLazyLoad(fsPath: string): boolean {
...
+    const createWorkspaceResponse = await fetch(
+        `${config.dataverseOrgUrl}/api/data/v9.2/sharedworkspaces`,
+        {
+            headers: {
+                ...config.headers,
+                Prefer: "return=representation",
+            },
+            method: "POST",
+            body: JSON.stringify({
+                name: config.websiteid,
```
tyaginidhi, line 261: Critical - this needs to follow same pattern as the studio container name - else we will have complete mis-match in read write data

src/web/client/utilities/urlBuilderUtil.ts
```diff
@@ -227,4 +227,42 @@ export function isWebFileWithLazyLoad(fsPath: string): boolean {
...
+                sharedworkspaceid: "00218f43-15d4-f87e-0e08-5dec2c4cfbaa",
```
tyaginidhi, line 262: generate random guid here - but mostly this is a no-op param here

tyaginidhi (changes requested):

src/web/client/WebExtensionContext.ts
ritikramuka: yes in my opinion as in further ms graph service implementation, we need to call this provider there also

src/web/client/context/userDataMap.ts
ritikramuka: address when implement active editor user list

src/web/client/extension.ts
ritikramuka: tried didn't worked! so keeping it as it is

src/web/client/utilities/urlBuilderUtil.ts
ritikramuka: addressed in latest commit

src/web/client/extension.ts
ritikramuka: this done

BidishaMS (changes requested):
Please break this PR

PR #697, opened 2023-09-14
Complete Navigation loop - add go to runtime and back to studio

tyaginidhi:
Adding back to studio and go to runtime options in vscode web:
image

NOTE: About the runtime API being used here (same API as is being used in Studio) - this is what the Runtime team mentioned "It will make the 1st request a bit slow, no other impact unless its being called too often". Given this API has not caused issues so far in studio, even with a higher volume - this should have a low impact on runtime performance. Have added telemetry to gauge number and behavior around this.


src/web/client/webViews/powerPagesNavigationProvider.ts
```diff
@@ -0,0 +1,148 @@
...
+        WebExtensionContext.dataverseAuthentication();
+
+        await vscode.window.withProgress(
+            {
+                location: vscode.ProgressLocation.Notification,
+                cancellable: true,
+                title: vscode.l10n.t("Opening preview site..."),
+            },
+            async () => {
+                const response = await WebExtensionContext.concurrencyHandler.handleRequest(requestUrl, {
```
tyaginidhi, line 84: call to runtime clear cache API

package.json
```diff
@@ -153,6 +153,14 @@
       }
     ],
     "commands": [
+      {
+        "command": "powerpages.powerPagesFileExplorer.powerPagesRuntimePreview",
```
amitjoshi438, line 157: Do we wish to show these commands in command palette also?

src/web/client/extension.ts
```diff
@@ -187,6 +191,13 @@ export function activate(context: vscode.ExtensionContext): void {
     showWalkthrough(context, WebExtensionContext.telemetry);
 }
 
+export function powerPagesNavigation() {
+    const powerPagesNavigationProvider = new PowerPagesNavigationProvider();
+    vscode.window.registerTreeDataProvider('powerpages.powerPagesFileExplorer', powerPagesNavigationProvider);
```
amitjoshi438, line 196: Is disposal of commands required in web extensions? Like use of context.subscriptions.push()

src/web/client/utilities/commonUtil.ts
```diff
@@ -199,10 +202,24 @@ export function isPortalVersionV2(): boolean {
...
+    return (WebExtensionContext.urlParametersMap.get(queryParameters.ENV_ID) as string).split("/")?.pop() as string;
+}
+
+export function getBackToStudioURL() {
+    const region = WebExtensionContext.urlParametersMap.get(queryParameters.REGION) as string;
+
+    return BACK_TO_STUDIO_URL_TEMPLATE
+        .replace("{environmentId}", getEnvironmentIdFromUrl())
+        .replace("{.region}", region.toLowerCase() === STUDIO_PROD_REGION ? "" : `.${WebExtensionContext.urlParametersMap.get(queryParameters.REGION) as string}`)
+        .replace("{webSiteId}", WebExtensionContext.urlParametersMap.get(queryParameters.WEBSITE_ID) as string);
```
BidishaMS, line 224: Is the expectation we always fallback to Homepage even though user navigates from other page.

src/web/client/webViews/powerPagesNavigationProvider.ts
```diff
@@ -0,0 +1,153 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
```
BidishaMS, line 3: Please break down into small PR. We can sync on this 
@ashishchoudhary001

BidishaMS (changes requested):

src/web/client/utilities/commonUtil.ts
tyaginidhi: that is correct

package.json
tyaginidhi: No for now - updating accordingly

src/web/client/extension.ts
tyaginidhi: Ideally we should - taking this as a separte task to be addressed in next PR.

amitjoshi438 (approved):

package.json
```diff
@@ -886,9 +896,12 @@
       ],
       "explorer": [
         {
-          "id": "powerpages.treeWebView",
-          "name": "Power Pages Actions",
-          "when": "isWeb && config.powerPlatform.experimental.enableCoPresenceFeature"
+          "id": "powerpages.powerPagesFileExplorer",
+          "name": "POWER PAGES ACTIONS",
```
tehcrashxor, line 900: Name should be localized like the `pacCLI.authPanel` name above.
Looks like we missed that for the "Copilot In Power Page" -- that should be localized as well

tehcrashxor (changes requested):

package.json
tyaginidhi: updated  - will ask Amit to address co-pilot one in a differen PR - if thats needed  -given we are only available in US geo - i beleieve that is intentional.

tehcrashxor (approved):

PR #702, opened 2023-09-15
userDataMap intreface for co-presence audience

ritikramuka:
image

containerId: string -> id of FRS container
fileName: string -> name of current active file
filePath: string -> path of current active file
userName: string -> name of user
userId: string -> user id

### DIFF
--- a/src/web/client/context/userDataMap.ts
+++ b/src/web/client/context/userDataMap.ts
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ */
+
+export interface IUserData {
+    containerId: string;
+    fileName: string;
+    filePath: string;
+    userName: string;
+    userId: string;
+}
+
+export class UserData implements IUserData {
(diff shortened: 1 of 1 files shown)


src/web/client/context/userDataMap.ts
```diff
@@ -0,0 +1,81 @@
...
+    _filePath: string;
+    _userName: string;
+    _userId: string;
+
+    // Getters
+    public get containerId(): string {
+        return this._containerId;
+    }
+    public get fileName(): string {
+        return this.fileName;
```
ashishchoudhary001, line 26: are you missing an underscore here?

src/web/client/context/userDataMap.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+    public get filePath(): string {
+        return this.filePath;
```
ashishchoudhary001, line 29: are you missing an underscore here?

src/web/client/context/userDataMap.ts
ritikramuka: updated

petrochuk (approved):

PR #703, opened 2023-09-15
Added co-presence worker script

ritikramuka:
added co-presence worker script to read Azure Fluid Client instance
- container
- audience
- userSharedMap


src/web/client/services/copresenceWorker.js
```diff
@@ -0,0 +1,165 @@
...
+                    userId: changed.key,
+                    userName: otherUser.userName,
+                    containerId: swpId,
+                    fileName: otherUser.fileName,
+                    filePath: otherUser.filePath,
+                });
+            }
+        });
+    } catch (error) {
+        WebExtensionContext.telemetry.sendErrorTelemetry(
```
tyaginidhi, line 143: confirm once if the telemetry form here actually works

ritikramuka: added as a TODO, as need some further investigation

ritikramuka:
closed this pull request! created a fresh with some better approach  https://github.com/microsoft/powerplatform-vscode/pull/731

PR #717, opened 2023-09-19
Logic to compile co-presence worker script

ritikramuka:
added logic to compile co-presence worker script in webpack config and gulp file


webpack.config.js
```diff
@@ -122,4 +122,58 @@ const webConfig = {
...
+/** @type fluent container scripts web worker config */
+const webWorkerConfig = {
+    mode: "none",
+    target: "webworker", // web extensions run in a webworker context
+    entry: {
+        main: "./src/web/client/common/worker/webworker.js",
+    },
+    output: {
+        filename: "[name].js",
+        path: path.join(__dirname, "./dist/web"),
```
ritikramuka, line 134: @amitjoshi438 this is the location where you suggested adding the webworker compiled file in the dist folder?

gulpfile.mjs
```diff
@@ -81,6 +81,14 @@ function compileWeb() {
         .pipe(gulp.dest(path.resolve(`${distdir}/web`)));
 }
 
+// compile the web worker for vscode for web
+function compileWorker() {
+    return gulp
+        .src(["src/web/**/*.ts"])
```
tyaginidhi, line 87: the worker path is now in common folder correct?

webpack.config.js
tyaginidhi: better keep it in dist/common ?

tyaginidhi: let it be in web for now

tehcrashxor (approved):

PR #727, opened 2023-09-29
[Copilot] Label for selected code

amitjoshi438:
- Shows the range of selected code in copilot panel
- It will be used to determine the code to be passed to LLM for **explain/fix** features

image


src/common/Utils.ts
```diff
@@ -7,94 +7,108 @@
...
+export function getSelectedCodeLineRange(editor: vscode.TextEditor): { start: number, end: number } {
+    if (!editor) {
+        return { start: 0, end: 0 };
+    }
+    // Get the selection(s) in the editor
+    const selection = editor.selection;
+
+    const startLine = selection.start.line;
+    const endLine = selection.end.line;
+    console.log(`Selected lines: ${startLine + 1}-${endLine + 1}`);
```
BidishaMS, line 28: We need console.log

src/common/Utils.ts
```diff
@@ -7,94 +7,108 @@
...
+    if (!editor) {
+        return "";
+    }
+    const selection = editor.selection;
+    const text = editor.document.getText(selection);
+    return text;
 }
 
+export function getSelectedCodeLineRange(editor: vscode.TextEditor): { start: number, end: number } {
+    if (!editor) {
```
BidishaMS, line 20: What you will return if the user selects empty space without any code?

src/common/Utils.ts
tyaginidhi, line 28: remove these console logs

tyaginidhi (approved):

tyaginidhi:
Also add more description for background - image is good - but for folks who dont know what this one is about - it wil give background - or maybe just highlight changes in the screenshot

src/common/Utils.ts
amitjoshi438: If the selected code length is non-zero, it will return the value for that line irrespective of content.

src/common/Utils.ts
amitjoshi438: removed

src/common/Utils.ts
amitjoshi438: Update: It now handles the empty selection and label will not be shown in such scenarios.

BidishaMS (approved):

tehcrashxor (approved):

PR #730, opened 2023-10-04
[Copilot] Show user selected code along with prompt in copilot panel

amitjoshi438:
- If user use **Explain** from editor context menu or enter some prompt after selecting code, it will be sent to LLM along with user prompt and will be shown in copilot panel also.

Note: Currently not passing the selected code to LLM in a separate field, will do that once changes on backend side are done.

image


src/common/copilot/assets/scripts/copilot.js
```diff
@@ -523,35 +505,36 @@
...
+  }
+
+
+  SendButton?.addEventListener("click", () => {
+    processUserInput(chatInput.value.trim());
   });
 
   chatInput.addEventListener("keydown", (event) => {
-    if(apiResponseInProgress) {
+    if(apiResponseInProgress && event.key !== "Enter") {
```
tyaginidhi, line 530: does casing for enter matter here?

tyaginidhi (approved):

src/common/copilot/assets/scripts/copilot.js
amitjoshi438: Yes, it's upper case by default

src/common/copilot/PowerPagesCopilot.ts
```diff
@@ -68,16 +68,28 @@ export class PowerPagesCopilot implements vscode.WebviewViewProvider {
...
     if (SELECTED_CODE_INFO_ENABLED) { //TODO: Remove this check once the feature is ready
+
+        const handleSelectionChange = async (commandType: string) => {
+            const editor = vscode.window.activeTextEditor;
+            if (!editor) {
+                return;
+            }
+            const selectedCode = getSelectedCode(editor);
+            const selectedCodeLineRange = getSelectedCodeLineRange(editor);
+            if(commandType === "explainCode" && selectedCode.length === 0) {
```
BidishaMS, line 79: can we make the text case-insensitive like both upper or lowercase and even trim spaces

src/common/copilot/assets/scripts/copilot.js
```diff
@@ -487,13 +461,21 @@
...
                 chatInputLabel.classList.add("hide");
                 break;
             }
             chatInputLabel.classList.remove("hide");
             chatInputLabel.innerText = `Lines: ${message.value.start + 1} - ${message.value.end + 1} selected`;
+            break;
+        }
+        case "explainCode": {
+            selectedCode = message.value.selectedCode;
+            const explainPrompt = 'Explain the following code:';
```
BidishaMS, line 475: Best practices. Should be read from a constant.

petrochuk, line 475: String needs to be localized

tehcrashxor: I'm not sure if the `.js` asset files can be, or perhaps just not easily with the existing vscode pseudo-module.

IIRC, Nidhi mentioned that copilot isn't yet available outside of US region and en-US language, so this might not need localization yet?

amitjoshi438: That's correct, current vscode localization module is not supported in webviews and currently copilot is only available in US region.

petrochuk: Even if it is not supported right now. It might be good idea to move strings to one location so localization later can be easily done

src/common/copilot/PowerPagesCopilot.ts
amitjoshi438: Added to constant, both the places same constant will be used. So, now we don't require trim and case handling.

src/common/copilot/assets/scripts/copilot.js
amitjoshi438: Right, will be moving all the string out of webview. So, they can be localized first and then sent to webview. Will take this up in separate as it will changes for all existing strings also.

amitjoshi438: Will be moving the strings out, currently there is no separate constant file for webview script.

petrochuk: It would be great to have it in one PR

amitjoshi438: Added

petrochuk (approved):

tehcrashxor (approved):

PR #731, opened 2023-10-04
co-presence worker script added

ritikramuka:
Co-Presence Worker Script, has a all the event listeners on FRS container to fetch and write back (coming soon) in fluid container.
it also get all the content entity id for a users via his all the connection id


amitjoshi438:
@ritikramuka can you please provide a generic name to worker script as it may be used for integrating other things apart from copresence. Also place it inside src/common folder and the webpack output should also be outside the /web folder.

amitjoshi438 (changes requested):
@ritikramuka can you please provide a generic name to worker script as it may be used for integrating other things apart from copresence. Also place it inside src/common folder and the webpack output should also be outside the /web folder.

ritikramuka:
> > @ritikramuka can you please provide a generic name to worker script as it may be used for integrating other things apart from copresence. Also place it inside src/common folder and the webpack output should also be outside the /web folder.
> 
> +1 on above request - we will be using the same worker for any web worker needed scripts to be run for web extension - so for now its copresence and copilot related stuff
1. what should be location of worker file? will we use it on desktop extension also... as @amitjoshi438 mentioned to place in src/common
2. I'm not managing webpack output in this PR, but will keep in mind in that PR... if I'm missing anything do let me know

src/web/client/common/worker/webworker.js
```diff
@@ -0,0 +1,165 @@
...
+
+        map.on("valueChanged", async (changed, local) => {
+            if (!local) {
+                const otherUser = map.get(changed.key);
+                const connectionArray = otherUser.connections;
+
+                const EntityID = [];
+
+                for (let i = 0; i < connectionArray.length; i++) {
+                    const connection = connectionArray[i]?.id;
```
tyaginidhi, line 132: better naming here - ConnectionId instead of connection

src/web/client/common/worker/webworker.js
```diff
@@ -0,0 +1,165 @@
...
+
+        if (!initial) {
+            existingMembers.forEach(async (value, key) => {
+                const otherUser = value;
+                const connectionArray = otherUser.connections;
+
+                const EntityID = [];
+
+                for (let i = 0; i < connectionArray.length; i++) {
+                    const connection = connectionArray[i]?.id;
```
tyaginidhi, line 109: variable name should be better - this is connectionId

src/web/client/common/worker/webworker.js
```diff
@@ -0,0 +1,165 @@
...
+}
+
+async function loadContainer(config, id, swpId) {
+    try {
+        const { container, audience, map } =
+            await AzureFluidClient.fetchContainerAndService(config, swpId);
+
+        const existingMembers = audience.getMembers();
+
+        // TODO: insert user with entity id in container
```
tyaginidhi, line 90: what is this about?

src/web/client/common/worker/webworker.js
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        map.on("valueChanged", async (changed, local) => {
+            if (!local) {
+                const otherUser = map.get(changed.key);
+                const connectionArray = otherUser.connections;
+
+                const EntityID = [];
+
+                for (let i = 0; i < connectionArray.length; i++) {
```
tyaginidhi, line 131: nit: use connectionArray.foreach instead?

src/web/client/common/worker/webworker.js
```diff
@@ -0,0 +1,165 @@
...
+                    const connection = connectionArray[i]?.id;
+                    EntityID.push((await container.initialObjects.sharedState.get('selection').get()).get(connection));
```
tyaginidhi, line 133: why await this explicitly every single time?

await container.initialObjects.sharedState.get('selection').get()

src/web/client/common/worker/webworker.js
```diff
@@ -0,0 +1,165 @@
...
+                    containerId: swpId,
+                    entityId: EntityID,
+                });
+            });
+            initial = true;
+        }
+
+        map.on("valueChanged", async (changed, local) => {
+            if (!local) {
+                const otherUser = map.get(changed.key);
```
amitjoshi438, line 126: can we use early return here? 
```
if (local) { 
return
}
```

amitjoshi438 (approved):

src/web/client/common/worker/webworker.js
ritikramuka: rn I'm not adding user entityId joined from vscode side! we need to have a discussion what will be the formate in which we will add it

src/web/client/common/worker/webworker.js
ritikramuka: done!

src/web/client/common/worker/webworker.js
ritikramuka: done!

src/web/client/common/worker/webworker.js
ritikramuka: okay!

src/web/client/common/worker/webworker.js
ritikramuka: yes! updated!

petrochuk (approved):

PR #732, opened 2023-10-05
Added filepath in entity data map

ritikramuka:
image
there can be multiple filepath against a entity id, (eg. home.html, home.css, home.js)


src/web/client/dal/remoteFetchProvider.ts
```diff
@@ -443,6 +443,14 @@ async function createFile(
         fileContent = GetFileContent(result, attributePath, entityName, entityId);
     }
 
+    const filePath : string[] = [];
+
+    Array.from(WebExtensionContext.fileDataMap.getFileMap.entries()).forEach(([fileUri, data]) => {
```
tyaginidhi, line 448: This does not seem right - what are you trying to do here?

ritikramuka: from file data map storing keys (file uri), in entity data map against its entity id

tyaginidhi: lets use the approch we discussed  - we dont need to do this iteration on a map

tyaginidhi (changes requested):

src/web/client/dal/remoteFetchProvider.ts
ritikramuka: updated

src/web/client/dal/remoteFetchProvider.ts
```diff
@@ -345,6 +346,10 @@ async function processDataAndCreateFile(
             let fileCreationValid = true;
             let fileNameWithExtension = GetFileNameWithExtension(entityName, fileName, languageCode, fileExtension);
 
+            const filePath = entityFileMap.get(entityId) ?? [];
+            filePath.push(fileNameWithExtension);
```
tyaginidhi, line 350: fileNameWithExtension is just the full file name right? Not the fileUri - check once

src/web/client/context/entityDataMap.ts
```diff
@@ -49,7 +50,8 @@ export class EntityDataMap {
             entityName,
             odataEtag,
             entityColumnMap,
-            mappingEntityId
+            mappingEntityId,
+            filePath
```
tyaginidhi, line 54: this will end up over-writing any previosuly saved filepaths for an entity like webpage

src/web/client/dal/remoteFetchProvider.ts
```diff
@@ -409,7 +415,8 @@ async function createFile(
     mappingEntityFetchQuery: string | undefined,
     entityId: string,
     dataverseOrgUrl: string,
-    portalsFS: PortalsFS
+    portalsFS: PortalsFS,
+    entityFileMap: Map<string, string[]>
```
tyaginidhi, line 419: why do we need this entityId map - at a time this call will come for just one entityId and the createVirtual file call will happen for just one of the URI (in case of webpage) associate with an entityId

src/web/client/dal/remoteFetchProvider.ts
ritikramuka: changing to fileuri

tyaginidhi (approved):

tehcrashxor (approved):

petrochuk (approved):

PR #733, opened 2023-10-06
Image file edit using luna paint extension for V1 and V2 data model

tyaginidhi:
We can now support edit of Image files with File extension - **png, jpg, jpeg, webp, bmp, tga and ico**

This is possible by using Luna Paint and other similar extensions available in Vscode for web.

Issue tracking the support related to this from Luna paint extension and vscode-dev is here for easy access and background on this:
vscode-dev: https://github.com/microsoft/vscode-dev/issues/988
Luna paint: https://github.com/lunapaint/vscode-luna-paint/issues/192


amitjoshi438 (approved):

src/web/client/utilities/commonUtil.ts
```diff
@@ -223,3 +224,43 @@ export function getBackToStudioURL() {
         .replace("{.region}", region.toLowerCase() === STUDIO_PROD_REGION ? "" : `.${WebExtensionContext.urlParametersMap.get(queryParameters.REGION) as string}`)
         .replace("{webSiteId}", WebExtensionContext.urlParametersMap.get(queryParameters.WEBSITE_ID) as string);
 }
+export function getSupportedImageFileExtensionsForEdit() {
+    return ['png', 'jpg', 'webp', 'bmp', 'tga', 'ico', 'jpeg']; // Luna paint supported image file extensions
+}
```
Tyriar, line 229: Why do you need an allow list for this? There are plenty of other files that may also be binary files that, for example the hex editor supports opening _any_ file extension to view it's data as hex?

tehcrashxor (approved):

src/web/client/utilities/commonUtil.ts
tyaginidhi: This allow list is just for write back to our backend - we allow reading of a file using any editor that vscode supports. Following up more on chat around this.

tyaginidhi: updating list per Luna paint supported extensions list.

PR #741, opened 2023-10-17
Use single extension with prerelease flag via pipeline parameter

tehcrashxor:
Move to using a single vscode extension -- use the pre-release flag instead of building an entirely separate preview extension.
Use Release-It to auto-increment the version number during Official Builds


tyaginidhi (approved):

.azure-pipelines/OfficialBuild.yml
```diff
@@ -63,16 +73,10 @@ steps:
...
-    command: custom
-    customCommand: run dist -- --feedPAT $(AZ_DevOps_Read_PAT) --isOfficialBuild true
-
-- task: Npm@1
-  displayName: 'Preview Build and Package VSIX'
+  displayName: 'Build and Package VSIX'
   inputs:
     command: custom
-    customCommand: run dist -- --feedPAT $(AZ_DevOps_Read_PAT) --isPreviewBuild true
+    customCommand: run dist -- --feedPAT $(AZ_DevOps_Read_PAT) --isOfficialBuild true --isPreviewBuild ${{ parameter.isPreRelease }}
```
tyaginidhi, line 79: we just need to make sure that isPreviewBuild default to true in our pipelines to avoid future automatic release pipelines leaking out anything to stable version.

petrochuk (approved):

.azure-pipelines/OfficialBuild.yml
tehcrashxor: The pre-release variable is set to true by default

PR #756, opened 2023-10-30
RootWebPageId in Entity Data Map for V1 and V2 data model

ritikramuka:
In this PR I'm trying to add rootpageId in the entity data

### DIFF
--- a/src/web/client/dal/remoteFetchProvider.ts
+++ b/src/web/client/dal/remoteFetchProvider.ts
@@ -271,6 +271,16 @@ async function createContentFiles(
         }
 
         const attributeArray = attributes.split(",");
+
+        // Get rootpage id
+        const content = entityDetails?.get(schemaEntityKey.CONTENT);
+
+        const contentValue = content ? result[content] : null;
+        if (contentValue === null) {
+            throw new Error(ERRORS.CONTENT_EMPTY);
+        }
+        const rootpageid = getRootPageId(contentValue);
+
         await processDataAndCreateFile(attributeArray,
             attributeExtension,
             entityName,
@@ -282,7 +292,8 @@ async function createContentFiles(
             languageCode,
             filePathInPortalFS,
             portalsFS,
-            defaultFileInfo)
+            defaultFileInfo,
+            rootpageid)
 
     } catch (error) {
         const errorMsg = (error as Error)?.message;
@@ -296,6 +307,11 @@ async function createContentFiles(
             error as Error
         );
     }
+}
+
+function getRootPageId(contentValue: string) : string {
+    const contentValueJson = JSON.parse(contentValue);
+    return contentValueJson.rootwebpageid;
 }
 
 async function processDataAndCreateFile(
@@ -312,6 +328,7 @@ async function processDataAndCreateFile(
     filePathInPortalFS: string,
     portalsFS: PortalsFS,
     defaultFileInfo?: IFileInfo,
+    rootpageid?: string
 ) {
     const attributeExtensionMap = attributeExtension as unknown as Map<
         string,
@@ -363,7 +380,8 @@ async function processDataAndCreateFile(
                     mappingEntityFetchQuery,
                     entityId,
                     dataverseOrgUrl,
-                    portalsFS
+                    portalsFS,
+                    rootpageid,
                 );
             }
         }
@@ -409,7 +427,8 @@ async function createFile(
     mappingEntityFetchQuery: string | undefined,
     entityId: string,
     dataverseOrgUrl: string,
-    portalsFS: PortalsFS
+    portalsFS: PortalsFS,
+    rootpageid?: string
 ) {
     const base64Encoded: boolean = isBase64Encoded(
         entityName,
@@ -458,7 +477,8 @@ async function createFile(
         mimeType ?? result[Constants.MIMETYPE],
         isPreloadedContent,
         mappingEntityId,
-        getLogicalEntityName(result, logicalEntityName)
+        getLogicalEntityName(result, logicalEntityName),
+        rootpageid,
     );
 }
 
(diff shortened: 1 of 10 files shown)


amitjoshi438 (approved):

petrochuk (approved):

src/web/client/schema/portalSchema.ts
```diff
@@ -307,6 +308,7 @@ export const portal_schema_V2 = {
                     ["content.customjavascript", "customjs.js"],
                     ["content.copy", "webpage.copy.html"],
                 ]),
+                _content: "content",
```
tyaginidhi, line 311: we are already getting content in call above - thats how we are fetching content.copy
DO not add this param here

src/web/client/schema/portalSchema.ts
```diff
@@ -80,13 +80,14 @@ export const portal_schema_V1 = {
...
                 _multiFileFetchQueryParameters:
-                    "?$filter=_adx_websiteid_value eq {websiteId} and _adx_webpagelanguageid_value ne null &$select=adx_webpageid,_adx_webpagelanguageid_value,adx_name,adx_copy,adx_customcss,adx_customjavascript,adx_partialurl&$count=true",
+                    "?$filter=_adx_websiteid_value eq {websiteId} and _adx_webpagelanguageid_value ne null &$select=adx_webpageid,_adx_webpagelanguageid_value,adx_name,adx_copy,adx_customcss,adx_customjavascript,adx_partialurl,_adx_rootwebpageid_value&$count=true",
                 _attributes: "adx_customcss,adx_customjavascript,adx_copy",
                 _attributesExtension: new Map([
                     ["adx_customcss", "customcss.css"],
                     ["adx_customjavascript", "customjs.js"],
                     ["adx_copy", "webpage.copy.html"],
                 ]),
+                _rootwebpageid: "_adx_rootwebpageid_value",
```
tyaginidhi, line 90: same for this change - we do not need to add _rootwebpageid here. Just fetch it in the fetchQueryparam calls and use it after reading this value from constants.

Also - you need to add it for both _fetchQueryParameters and _multiFileFetchQueryParameters. We support single file and multi file view of files still.

src/web/client/schema/constants.ts
```diff
@@ -30,6 +30,8 @@ export enum schemaEntityKey {
     MAPPING_ENTITY_FETCH_QUERY = "_mappingEntityFetchQuery",
     EXPORT_TYPE = "_exporttype",
     ATTRIBUTES = "_attributes",
+    CONTENT = "_content",
+    ROOT_WEB_PAGE_ID = "_rootwebpageid",
```
tyaginidhi, line 34: this is the only place you need this constant and not below. Add _adx_rootwebpageid_value for V1 and content.rootwebpageid here.

src/web/client/dal/remoteFetchProvider.ts
```diff
@@ -298,6 +324,11 @@ async function createContentFiles(
     }
 }
 
+function getRootPageId(contentValue: string) : string {
```
tyaginidhi, line 327: You do not need this function. Use GetFileContent() in commonUtil file. You can use getAttributePath() function to convert V1 and V2 field values to this format and then GetFileContent will do the rest for you. You dont need to do V1 and V2 specific code handling in this fetchProvider at all.

src/web/client/dal/remoteFetchProvider.ts
```diff
@@ -271,6 +271,31 @@ async function createContentFiles(
         }
 
         const attributeArray = attributes.split(",");
+
+        // Get rootpage id
+        let rootWebPageId = undefined;
+
+        if (isPortalVersionV2()) {
```
tyaginidhi, line 278: Refer comment below - you dont need V1 and V2 specific handling here.

src/web/client/dal/remoteFetchProvider.ts
```diff
@@ -271,6 +271,31 @@ async function createContentFiles(
...
+                }
+                rootWebPageId = getRootPageId(contentValue);
+            }
+        } else if (isPortalVersionV1()) {
+            const rootpageidAttribute = entityDetails?.get(schemaEntityKey.ROOT_WEB_PAGE_ID);
+
+            if (rootpageidAttribute) {
+                rootWebPageId = rootpageidAttribute ? result[rootpageidAttribute] : null;
+                if (rootWebPageId === null) {
+                    throw new Error(ERRORS.ROOT_WEB_PAGE_ID_EMPTY);
```
tyaginidhi, line 294: why do you want to block the whole page load if rootwebPageId is missing - though this will be conver most case and other things will be broken at this point - but Rule of Thumb for copresence specific code is dont block other flows. Just log an error. And make sure copresence code handles any missing data scenario.

tyaginidhi (changes requested):

src/web/client/schema/portalSchema.ts
ritikramuka: _rootwebpageid is added to get the attribute name, didn't want to make it confusing by adding with _attributes field, which has file extension

src/web/client/utilities/commonUtil.ts
```diff
@@ -264,3 +264,22 @@ export function getImageFileContent(fileFsPath: string, fileContent: Uint8Array)
 
     return webFileV2 ? fileContent : Buffer.from(fileContent).toString(BASE_64);
 }
+
+export function getRootWebPageId(result: any, attributePath: IAttributePath, entityName: string, entityId: string) {
+    let rootWebPageId = result[attributePath.source] ?? NO_CONTENT;
```
tyaginidhi, line 269: reuse existing function for this?

tyaginidhi (approved):

PR #761, opened 2023-10-31
Added artemis response (geo) on extension load

amitjoshi438:


src/web/client/extension.ts
```diff
@@ -459,3 +462,19 @@ function isActiveDocument(fileFsPath: string): boolean {
         WebExtensionContext.fileDataMap.getFileMap.has(fileFsPath)
     );
 }
+
+async function logArtemisTelemetry() {
+    const orgId =  WebExtensionContext.urlParametersMap.get(
```
tyaginidhi, line 467: add a try catch on this call - we dont want the extension to fail loading in case this function fails.

tyaginidhi (approved):

BidishaMS (approved):

src/web/client/extension.ts
```diff
@@ -460,3 +463,28 @@ function isActiveDocument(fileFsPath: string): boolean {
...
+        }
+
+        const { geoName } = artemisResponse[0];
+        WebExtensionContext.telemetry.sendInfoTelemetry(telemetryEventNames.WEB_EXTENSION_ARTEMIS_RESPONSE,
+            { orgId: orgId, geoName: String(geoName) });
+    } catch (error) {
+        WebExtensionContext.telemetry.sendErrorTelemetry(
+            telemetryEventNames.WEB_EXTENSION_ARTEMIS_RESPONSE_FAILED,
+            logArtemisTelemetry.name,
+            (error as Error)?.message,
```
petrochuk, line 487: Please don't use telemetry as replacement for logging

petrochuk (changes requested):

src/web/client/extension.ts
ashishchoudhary001: This code runs in VSCode Web. Here we don't have the option to log in PAC Log file.

petrochuk (approved):

PR #769, opened 2023-11-10
graph client intergration

ritikramuka:
this PR contains:
- Microsoft Graph Client Authentication
- Calls for different services from the Microsoft Graph
    - email
    - profile photo


src/web/client/utilities/graphClientUtils.ts
```diff
@@ -0,0 +1,85 @@
...
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ */
+
+import path from "path";
+import WebExtensionContext from "../WebExtensionContext";
+import { getCommonHeaders } from "../common/authenticationProvider";
+import * as Constants from "../common/constants";
+import { telemetryEventNames } from "../telemetry/constants";
+
+export async function requestGraphClient(graphToken: string, service: string, userId: string) {
```
tyaginidhi, line 12: this does not need to be an export function since all graph calls you will make will be added to this file.

src/web/client/utilities/graphClientUtils.ts
```diff
@@ -0,0 +1,85 @@
...
+    let requestSentAtTime = new Date().getTime();
```
tyaginidhi, line 13: one recommendation is to create a class for all these graph functions - refer telemetry class example. You can initialize that class in webExtensionContext and store the graph token in that class as private. Only functions inside that class will then access the token to make graph calls - and the whole extension wont have acccess to this token.

src/web/client/utilities/graphClientUtils.ts
tyaginidhi, line 12: this function can be names better - you are making a network call here and fetching a repsonse. The name suggests this is a call to get client - that will be used futher for subsequent flow.

src/web/client/utilities/graphClientUtils.ts
```diff
@@ -0,0 +1,85 @@
...
+            }
+        );
+
+        if(!response.ok) {
+            throw new Error(JSON.stringify(response));
+        }
+
+        WebExtensionContext.telemetry.sendAPISuccessTelemetry(
+            requestUrl.href,
+            "",
```
tyaginidhi, line 56: do not leave this empty - you need to provide some entity name here. For graph you can add a const in telemetry file and pass something like "GraphMail" or "GraphProfile" to distinguish two API call for performance analysis later

src/web/client/utilities/graphClientUtils.ts
```diff
@@ -0,0 +1,85 @@
...
+            requestGraphClient.name
+        );
+
+        return await response.json();
+    } catch (error) {
+        const errorMsg = (error as Error)?.message;
+        if ((error as Response)?.status > 0) {
+            WebExtensionContext.telemetry.sendAPIFailureTelemetry(
+                requestUrl.href,
+                "",
```
tyaginidhi, line 68: same here

src/web/client/WebExtensionContext.ts
```diff
@@ -645,6 +657,47 @@ class WebExtensionContext implements IWebExtensionContext {
...
+    }
+
+    public setGraphToken(graphToken: string) {
+        this._graphToken = graphToken;
+    }
+
+    private async getUserEmail(userId: string) {
+        try {
+            const response = await requestGraphClient(this.graphToken, Constants.GraphClientService.MAIL, userId);
+            return await response.mail;
```
tyaginidhi, line 674: suggestion is to log specific errors here and not in the "requestGraphCLient"function - that way you can define the "entityName" differently for mail and profile pic call

src/web/client/WebExtensionContext.ts
```diff
@@ -103,7 +108,8 @@ class WebExtensionContext implements IWebExtensionContext {
     private _npsEligibility: boolean;
     private _userId: string;
     private _formsProEligibilityId: string;
-    private _concurrencyHandler: ConcurrencyHandler
+    private _concurrencyHandler: ConcurrencyHandler;
+    private _graphToken: string;
```
tyaginidhi, line 112: If you make a class for graph calls - you will not need to store token here.

src/web/client/WebExtensionContext.ts
```diff
@@ -645,6 +657,47 @@ class WebExtensionContext implements IWebExtensionContext {
...
     }
+
+    // Graph client
+    public async graphClientAuthentication(firstTimeAuth = false) {
+        const accessToken = await graphClientAuthentication(firstTimeAuth);
+        this.setGraphToken(accessToken);
+    }
+
+    public setGraphToken(graphToken: string) {
+        this._graphToken = graphToken;
```
tyaginidhi, line 668: no need for this function in any case - we just update the private variable as this._variable = value

tyaginidhi (changes requested):

src/web/client/WebExtensionContext.ts
ritikramuka: image
logging service name now in requestGraphClient, does this address this?

src/web/client/WebExtensionContext.ts
```diff
@@ -189,7 +189,6 @@ class WebExtensionContext implements IWebExtensionContext {
     public get concurrencyHandler() {
         return this._concurrencyHandler;
     }
-
     constructor() {
```
tyaginidhi, line 192: nit: roll back this removal of line spacing before constructor

src/web/client/common/authenticationProvider.ts
```diff
@@ -161,3 +161,62 @@ export async function npsAuthentication(
...
+                `User.ReadBasic.All`,
+            ],
+            { silent: true }
+        );
+
+        if (!session) {
+            session = await vscode.authentication.getSession(
+                PROVIDER_ID,
+                [
+                    `Contacts.Read`,
```
tyaginidhi, line 183: save these as constants and consume here - similar to constants we use in dataverseAuthentication call

src/web/client/common/constants.ts
```diff
@@ -98,3 +98,12 @@ export enum portalSchemaVersion {
     V1 = "portalschemav1",
     V2 = "portalschemav2",
 }
+
+export const MICROSOFT_GRAPH_USERS_BASE_URL = "https://graph.microsoft.com/v1.0/users/";
+
+export enum GraphService {
```
tyaginidhi, line 104: The naming is not clear - rename this to something like GraphUsersProperties

src/web/client/services/graphClientService.ts
```diff
@@ -0,0 +1,152 @@
...
+        this._graphToken = "";
+    }
+
+    get graphToken() {
+        return this._graphToken;
+    }
+
+    public async graphClientAuthentication(firstTimeAuth = false) {
+        const accessToken = await graphClientAuthentication(firstTimeAuth);
+        this._graphToken = accessToken;
```
tyaginidhi, line 25: only set this if you have a non-empty token from auth call

src/web/client/services/graphClientService.ts
```diff
@@ -0,0 +1,152 @@
...
+            WebExtensionContext.telemetry.sendAPITelemetry(
+                requestUrl.href,
+                service,
+                Constants.httpMethod.GET,
+                this.requestGraphClient.name
+            );
+
+            requestSentAtTime = new Date().getTime();
+
+            const response =
```
tyaginidhi, line 64: only make this call is the graphToken is non-empty

src/web/client/services/graphClientService.ts
```diff
@@ -0,0 +1,152 @@
...
+        try {
+            if (!this._graphToken) {
+                await this.graphClientAuthentication(true);
+            }
+
+            const response = await this.requestGraphClient(
+                Constants.GraphService.GRAPH_MAIL,
+                userId
+            );
+            return await response.mail;
```
tyaginidhi, line 121: expect this to  be empty string in case of graph auth failures or API call faiures - we need to hook to the right fallback experience in that case

src/web/client/services/graphClientService.ts
```diff
@@ -0,0 +1,152 @@
...
+                    errorMsg,
+                    "",
+                    (error as Response)?.status.toString()
+                );
+            } else {
+                WebExtensionContext.telemetry.sendErrorTelemetry(
+                    telemetryEventNames.WEB_EXTENSION_GET_FROM_GRAPH_CLIENT_FAILED,
+                    this.requestGraphClient.name,
+                    (error as Error)?.message,
+                    error as Error
```
tyaginidhi, line 105: In light of recent recommendations - is there an alternate logging we need to do here?

tyaginidhi (approved):

tyaginidhi:
Approving this since expectation is graphClient error scenarios on auth or API failures will be addressed here - or as part of the next PR, correct?

ritikramuka:
> Approving this since expectation is graphClient error scenarios on auth or API failures will be addressed here - or as part of the next PR, correct?

will address the above comments here! and can see into some edge cases in subsequent PR

src/web/client/services/graphClientService.ts
ritikramuka: updated

tehcrashxor (approved):

PR #774, opened 2023-11-20
Quick Pick for Present Users on Studio or vscode side

ritikramuka:
In this PR we have a user icon to show all the users present, in a quick pick which is invoked via an action button
image

When no user present on the page
image


src/web/client/extension.ts
```diff
@@ -265,6 +265,7 @@ export function processWorkspaceStateChanges(context: vscode.ExtensionContext) {
                                     entityInfo
                                 });
                             }
+                            WebExtensionContext.quickPickProvider.updateQuickPickItems(entityInfo);
```
tyaginidhi, line 268: please move this entire logic under "if (isCoPresenceEnabled()) {" to its own function. This funciton is getting way to large .

tyaginidhi (approved):

tehcrashxor (approved):
Minor complaint: QuickPicks are intended to be input controls, not an output display.  
However there aren't a lot of options given to us by the VS Code API for better alternatives.   You could use a [MessageItem (showInformationMessage)](https://code.visualstudio.com/api/references/vscode-api#MessageItem) or an [OutputChannel](https://code.visualstudio.com/api/references/vscode-api#OutputChannel), but admittedly are clunkier and aren't great ways of displaying that information either.  ¯\\\_(ツ)_/¯

Approving the PR anway, as there's not much in the way of better options.

PR #814, opened 2024-01-18
Remove noisy notifications from fetch flow

tyaginidhi:


src/web/client/dal/remoteFetchProvider.ts
```diff
@@ -291,9 +287,7 @@ async function createContentFiles(
 
     } catch (error) {
         const errorMsg = (error as Error)?.message;
-        vscode.window.showErrorMessage(
-            vscode.l10n.t("Failed to get file ready for edit.")
-        );
+        console.error(vscode.l10n.t("Failed to get file ready for edit: {0}", fileName));
```
ritikramuka, line 290: can we have condition here, to console error when file name is not empty, as we are sending it empty name intentionally?

ritikramuka (approved):

amitjoshi438 (approved):

src/web/client/dal/remoteFetchProvider.ts
tyaginidhi: We need to have some indication to users that is not noisy (like notifications) and has some indication around what is wrong - even if filename is empty.

tehcrashxor (approved):

PR #826, opened 2024-02-05
Init telemetry logger for desktop extension

gshivi:
**Description**
This pull request implements modification to initialize the oneDSLogger for telemetry upon the activation of the desktop extension. The initialization is based on the orgId retrieved from the pac command(active organization), and the corresponding region is obtained from this orgId. These region details are utilized to start the logger.
Initially, the logger is set up with the region-us, and it can be dynamically updated later based on organizational details.

_How is active org call working?_
On extension initialization, an instance of OrgChangeNotifier class is created from activation function. This is a singleton class.
In its constructor - it makes a call for org details. 
Whenever call for activeOrg() pac call is success, a custom event of fired with org details.
Both copilot & extension subscribe to this event & further consume the value of org details & carry out further actions.
On changes to org - only this class will make the pac call for updated details.
Note: OrgChangeErrorEventEmitter is not throwing error details for now as the subscribers don't need those details. This can be updated in future as per need.


src/client/extension.ts
```diff
@@ -61,9 +66,14 @@ export async function activate(
         appInsightsResource.instrumentationKey
     );
     context.subscriptions.push(_telemetry);
+    oneDSLoggerWrapper.instantiate("us");
```
BidishaMS, line 69: add a TODO.

// TODO: Need to decide on data residency part on unauthenticated scenario

src/client/extension.ts
```diff
@@ -61,9 +66,14 @@ export async function activate(
...
+
     _telemetry.sendTelemetryEvent("Start", {
         "pac.userId": readUserSettings().uniqueId,
     });
+    oneDSLoggerWrapper.getLogger().traceInfo("Start", {
```
BidishaMS, line 74: Lets put the whole integration in a different PR. So as of now remove it

src/client/extension.ts
```diff
@@ -95,6 +105,9 @@ export async function activate(
                 _telemetry.sendTelemetryEvent("StartCommand", {
                     commandId: "microsoft-powerapps-portals.preview-show",
                 });
+                oneDSLoggerWrapper.getLogger().traceInfo("StartCommand", {
```
BidishaMS, line 108: same here

src/client/extension.ts
BidishaMS: Do it for the entire file

BidishaMS (approved):

src/client/extension.ts
```diff
@@ -192,13 +222,41 @@ export async function activate(
...
+    const pacContext = new PacWrapperContext(_context, _telemetry);
+    const interop = new PacInterop(pacContext, cliPath);
+    const pacWrapper = new PacWrapper(pacContext, interop);
+    const pacActiveOrg = await pacWrapper?.activeOrg();
+    if (pacActiveOrg && pacActiveOrg.Status === PAC_SUCCESS) {
+        const orgID = pacActiveOrg.Results.OrgId;
+        const artemisResponse = await fetchArtemisResponse(orgID, _telemetry);
+        if (artemisResponse) {
+            const { geoName } = artemisResponse[0];
+            oneDSLoggerWrapper.instantiate(geoName);
```
BidishaMS, line 251: Have you tested by changing geo. Is the data flowing to respective Aria clusters?

gshivi: On creating new project in Aria, data isn't flowing - so can't test.

gshivi: have created a tracking item: https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/3821881
Will take it up here.

tyaginidhi (approved):

src/client/extension.ts
```diff
@@ -192,13 +205,39 @@ export async function activate(
...
+        }
+    }
```
tehcrashxor, line 235: Duplicated code from `src/common/copilot/PowerPagesCopilot.ts`. 
- Filewatcher can be limited by the OS, so we should generally not have duplicates for the same file
- PAC calls are generally slow (and occur serially in a queue) so we should not be multiple calls for the same information

gshivi: Abstracted file watcher to a common re-usable function

gshivi: Create this tracking item related to optimizing activeOrg PAC call.
[https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/3830140](https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/3830140)

gshivi: Changes for activeOrg call optimization have been pushed to this PR itself.

amitjoshi438 (changes requested):

src/client/extension.ts
```diff
@@ -173,8 +193,12 @@ export async function activate(
             telemetryData = JSON.stringify(listOfActivePortals);
             _telemetry.sendTelemetryEvent("VscodeDesktopUsage", { listOfActivePortals: telemetryData, countOfActivePortals: listOfActivePortals.length.toString() });
         } catch (exception) {
-            _telemetry.sendTelemetryException(exception as Error, { eventName: 'VscodeDesktopUsage' });
+            const exceptionError = exception as Error;
+            _telemetry.sendTelemetryException(exceptionError, { eventName: 'VscodeDesktopUsage' });
         }
+        // Init OrgChangeNotifier instance
+        OrgChangeNotifier.createOrgChangeNotifierInstance(pacTerminal.getWrapper());
```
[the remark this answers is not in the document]
gshivi: Yes - for power pages
cc: @BidishaMS

src/common/copilot/PowerPagesCopilot.ts
```diff
@@ -104,10 +104,9 @@ export class PowerPagesCopilot implements vscode.WebviewViewProvider {
                 vscode.commands.registerCommand("powerpages.copilot.explain", () => { sendTelemetryEvent(this.telemetry, { eventName: CopilotExplainCode, copilotSessionId: sessionID, orgId: orgID }); this.show(); handleSelectionChange(EXPLAIN_CODE) })
             );
         }
-
-        if (this._pacWrapper) {
-            this.setupFileWatcher();
-        }
+        this._disposables.push(
```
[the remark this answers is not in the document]
gshivi: The OrgChangeNotifier class gets instantiated for desktop extension only. From a method in that class this event will be fired & then consumed here. Given, that class won' fire if web extension - this shouldn't get triggered as per my understanding.

src/common/OrgChangeNotifier.ts
```diff
@@ -0,0 +1,53 @@
...
+            const watcher = vscode.workspace.createFileSystemWatcher(watchPath);
+            watcher.onDidChange(() => this.activeOrgDetails());
+            watcher.onDidCreate(() => this.activeOrgDetails());
+            watcher.onDidDelete(() => this.activeOrgDetails());
+        }
+    }
+
+    private async activeOrgDetails() {
+        const pacActiveOrg = await this._pacWrapper?.activeOrg();
+        if (pacActiveOrg && pacActiveOrg.Status === PAC_SUCCESS) {
```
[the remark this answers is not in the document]
gshivi: Updated, will fire an even in case of failure. It can be subscribed where failure needs to be handled.

src/common/copilot/PowerPagesCopilot.ts
```diff
@@ -104,10 +104,9 @@ export class PowerPagesCopilot implements vscode.WebviewViewProvider {
...
+            orgChangeEvent((orgDetails: ActiveOrgOutput) => this.handleOrgChangeSuccess(orgDetails))
```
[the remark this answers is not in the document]
gshivi: Nice catch!  Have updated this. Also, removed redundant checks in case of failure from handleOrgChange method.

tehcrashxor (approved):

amitjoshi438 (approved):

PR #829, opened 2024-02-06
Update package.json with for security compliance issues

ritikramuka:
Resolve dependency conflicts and update vulnerable packages.

This commit addresses dependency conflicts identified by Dependabot and updates the project's dependencies to resolve vulnerabilities.

Alert 1: https://github.com/microsoft/powerplatform-vscode/security/dependabot/35

Changes made:
- Updated fluid-framework version: "^1.3.3" which is compatible with "@fluidframework/azure-client": "^1.1.1",

Unresolved:

**"@fluidframework/azure-client does not have a latest released version with updated axios version"**

Alter 2: https://github.com/microsoft/powerplatform-vscode/security/dependabot/32

### DIFF
--- a/package.json
+++ b/package.json
@@ -1112,9 +1112,5 @@
   "optionalDependencies": {
     "bufferutil": "^4.0.6",
     "utf-8-validate": "^5.0.9"
-  },
-  "resolutions": {
-    "axios": "0.21.4",
-    "jsrsasign": "11.0.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -1112,5 +1112,9 @@
   "optionalDependencies": {
     "bufferutil": "^4.0.6",
     "utf-8-validate": "^5.0.9"
+  },
+  "resolutions": {
+    "axios": "0.21.4",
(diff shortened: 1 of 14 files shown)


tehcrashxor (approved):

petrochuk (approved):

package.json
```diff
@@ -1085,7 +1085,7 @@
     "cockatiel": "^3.1.1",
     "command-exists": "^1.2.9",
     "find-process": "^1.4.7",
-    "fluid-framework": "^1.3.7",
+    "fluid-framework": "^1.3.3",
```
tehcrashxor, line 1088: Hmm.   Taking a second look, both of the alerts listed in the summary are complaining about transitive dependencies of `@fluidframework/azure-client`, not `fluid-framework`.  Would this change actually have any effect on those alerts?

ritikramuka: I have updated the description, so alert 1 is solved by this version change, can have a final verification after merging to main, and about the second alert currently "@fluidframework/azure-client does not have a latest released version with updated axios version", hence it is a known issue.

PR #830, opened 2024-02-06
Integrate OneDSLoggerWrapper in WebExtensionTelemetry

BidishaMS:
Integrate OneDSLoggerWrapper in WebExtensionTelemetry
Introduce a constant shortNameMappingtoAzureRegion. It will later used in OneDSLogger.

### DIFF
--- a/src/web/client/extension.ts
+++ b/src/web/client/extension.ts
@@ -46,7 +46,7 @@ export function activate(context: vscode.ExtensionContext): void {
         vscodeExtAppInsightsResourceProvider.GetAppInsightsResourceForDataBoundary(
             dataBoundary
         );
-    // Unauthenticated scenarios
+    // TODO: Need to replace the geo on confirmation from our Privacy champ. Unauthenticated scenarios
     oneDSLoggerWrapper.instantiate();
     WebExtensionContext.setVscodeWorkspaceState(context.workspaceState);
     WebExtensionContext.telemetry.setTelemetryReporter(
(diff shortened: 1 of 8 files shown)


src/web/client/test/integration/AuthenticationProvider.test.ts
```diff
@@ -67,6 +68,11 @@ describe("Authentication Provider", () => {
             "sendErrorTelemetry"
         );
 
+        const sendErrorTelemetry1 = sinon.spy(
```
ashishchoudhary001, line 71: Please rename the const to make it descriptive.

src/web/client/extension.ts
```diff
@@ -89,7 +92,11 @@ export function activate(context: vscode.ExtensionContext): void {
                         );
                     }
                 }
-
+                const geo = queryParamsMap.get('geo')?.toUpperCase();
+                // Authenticated scenario. Pass the geo to OneDSLogger for data boundary
+                if(geo !== undefined && geo !== null && geo !== ''){
```
ashishchoudhary001, line 97: instead of these checks can we replace it by simply checking if geo has a **truthy** value?

tyaginidhi (approved):

src/web/client/test/integration/AuthenticationProvider.test.ts
BidishaMS: fixed

src/web/client/extension.ts
BidishaMS: fixed

ashishchoudhary001 (approved):

ashishchoudhary001 (approved):

tehcrashxor (approved):

PR #839, opened 2024-02-09
Adding telemetry for entity and entityid in query params

ritikramuka:
This pull request add entity and entityid in telemetry depending on its presence in path url or query params


src/web/client/telemetry/webExtensionTelemetry.ts
```diff
@@ -29,10 +29,14 @@ export class WebExtensionTelemetry {
             eventName: telemetryEventNames.WEB_EXTENSION_INIT_PATH_PARAMETERS,
             properties: {
                 appName: this.getPathParameterValue(appName),
-                entity: this.getPathParameterValue(entity),
-                entityId: this.getPathParameterValue(entityId)
             }
         }
+
+        if (entity && entityId) {
```
tyaginidhi, line 35: do we really need both to be present? Just entity present is a scenario where you can load all the files for a single entitiy like for exampe "constent snippets" - right?

tyaginidhi (approved):

BidishaMS (approved):

src/web/client/telemetry/webExtensionTelemetry.ts
ritikramuka: The above scenario you are talking about might be broken or is not implemented in that way expected. Lets fix this in a subsequent task.

PR #849, opened 2024-02-18
Introducing Feature flag for telemetry in Power Pages

gshivi:
[https://code.visualstudio.com/api/extension-guides/telemetry](https://code.visualstudio.com/api/extension-guides/telemetry)

Based on users selection, telemetry will be logged.
By default, flag value is true.

### DIFF
--- a/src/common/OneDSLoggerTelemetry/oneDSLoggerWrapper.ts
+++ b/src/common/OneDSLoggerTelemetry/oneDSLoggerWrapper.ts
@@ -3,6 +3,7 @@
  * Licensed under the MIT License. See License.txt in the project root for license information.
  */
 
+import { isCustomTelemetryEnabled } from "../Utils";
 import { OneDSLogger } from "./oneDSLogger";
 
 //// Wrapper class of oneDSLogger for below purposes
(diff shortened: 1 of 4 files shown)


src/common/OneDSLoggerTelemetry/oneDSLoggerWrapper.ts
```diff
@@ -29,6 +30,7 @@ export class oneDSLoggerWrapper {
 	/// Trace info log
 	public traceInfo(eventName:string, eventInfo?: object, measurement?: object) {
         try{
+            if (!isCustomTelemetryEnabled()) return;
```
BidishaMS, line 33: We should not even initialize 1DSLogger instance

BidishaMS (changes requested):

BidishaMS (approved):

src/common/OneDSLoggerTelemetry/oneDSLoggerWrapper.ts
gshivi: Initializing would help handle the scenarios where users midway change their decision & in middle of work-session, enable/disable their telemetry setting. This controls custom telemetry logging - so whenever this would be false, telemetry won't be logged serving our use-case.

amitjoshi438 (approved):

PR #851, opened 2024-02-20
Handle orgGeo params for telemetry data residency

BidishaMS:
We are determining org geo using Artemis service. Artemis service gives us the org geo based on orgId and use the same geo for data residency

### DIFF
--- a/src/web/client/common/constants.ts
+++ b/src/web/client/common/constants.ts
@@ -98,7 +98,7 @@ export enum queryParameters {
     ORG_URL = "orgurl",
     REGION = "region",
     ENV_ID = "envid",
-    GEO = "geo",
+    GEO = "geo", // User geo location
     ENABLE_MULTIFILE = "enablemultifile",
     WEBSITE_PREVIEW_URL = "websitepreviewurl",
     ENTITY = "entity",
--- a/src/web/client/common/constants.ts
+++ b/src/web/client/common/constants.ts
@@ -103,8 +103,7 @@ export enum queryParameters {
     WEBSITE_PREVIEW_URL = "websitepreviewurl",
     ENTITY = "entity",
     ENTITY_ID = "entityid",
-    REFERRER_SOURCE = "referrersource",
-    ORG_GEO = 'orgGeo'
+    REFERRER_SOURCE = "referrersource"
 }
 
 export enum httpMethod {
--- a/src/web/client/common/constants.ts
+++ b/src/web/client/common/constants.ts
@@ -103,7 +103,8 @@ export enum queryParameters {
     WEBSITE_PREVIEW_URL = "websitepreviewurl",
     ENTITY = "entity",
     ENTITY_ID = "entityid",
-    REFERRER_SOURCE = "referrersource"
+    REFERRER_SOURCE = "referrersource",
+    ORG_GEO = 'orgGeo'
 }
 
 export enum httpMethod {
--- a/src/web/client/extension.ts
+++ b/src/web/client/extension.ts
@@ -92,12 +92,14 @@ export function activate(context: vscode.ExtensionContext): void {
                         );
                     }
                 }
-                const geo = queryParamsMap.get(queryParameters.ORG_GEO)?.toLowerCase() ?? 'us';
+                let geo = queryParamsMap.get(queryParameters.ORG_GEO)?.toLowerCase();
                 // Authenticated scenario. Pass the geo to OneDSLogger for data boundary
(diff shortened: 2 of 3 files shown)


src/web/client/extension.ts
```diff
@@ -92,7 +92,7 @@ export function activate(context: vscode.ExtensionContext): void {
                         );
                     }
                 }
-                const geo = queryParamsMap.get('geo')?.toLowerCase();
+                const geo = queryParamsMap.get(queryParameters.ORG_GEO)?.toLowerCase() ?? 'us';
```
tyaginidhi, line 95: is defaulting to US a good idea here? Specially given we can have EU org here?

src/web/client/common/constants.ts
```diff
@@ -103,7 +103,8 @@ export enum queryParameters {
     WEBSITE_PREVIEW_URL = "websitepreviewurl",
     ENTITY = "entity",
     ENTITY_ID = "entityid",
-    REFERRER_SOURCE = "referrersource"
+    REFERRER_SOURCE = "referrersource",
+    ORG_GEO = 'orgGeo'
```
tyaginidhi, line 107: lets leave a comment here that "geo" above means user geo - we shouldnt rename that param but lets leave a clarifying comment.

tyaginidhi (approved):

src/web/client/extension.ts
BidishaMS: I will check once with Dileep on this but this will act as unauthenticated scenario incase orgGeo is empty. As of now we have decided for all unauthenticated scenario will pass the logging to US

ashishchoudhary001, line 95: 1. Given this field determines where the telemetry data will be logged, I don't think we should have a fallback option here.
2. We should make 'geo' as a mandatory parameter.
3. We should think if this is the correct way to solve the problem of determining the cluster for logging telemetry data. A user can easily modify the url parameter and change the value of ORG_GEO in which case we would end up logging data in an incorrect cluster. I think we should leverage a service from VSCode e.g., Artemis to determine this info.

cc: @tyaginidhi

ashishchoudhary001 (changes requested):

src/web/client/extension.ts
amitjoshi438: We already have Artemis integration for vscode.dev, we can leverage that here.

amitjoshi438 (approved):

src/web/client/extension.ts
BidishaMS: @ashishchoudhary001 Called Artemis service to get the GeoName

BidishaMS: Integrated Artemis service response

src/web/client/common/constants.ts
```diff
@@ -98,12 +98,13 @@ export enum queryParameters {
...
     ENV_ID = "envid",
-    GEO = "geo",
+    GEO = "geo", // User geo location
     ENABLE_MULTIFILE = "enablemultifile",
     WEBSITE_PREVIEW_URL = "websitepreviewurl",
     ENTITY = "entity",
     ENTITY_ID = "entityid",
-    REFERRER_SOURCE = "referrersource"
+    REFERRER_SOURCE = "referrersource",
+    ORG_GEO = 'orgGeo'
```
tyaginidhi, line 107: with Artemis changes do we really need this change now? Lets not do both

BidishaMS: Make sense. Removed

src/web/client/extension.ts
```diff
@@ -601,20 +598,24 @@ function isActiveDocument(fileFsPath: string): boolean {
     );
 }
 
+async function fetchArtemisData(orgId: string) : Promise<string> {
+        const artemisResponse = await fetchArtemisResponse(orgId, WebExtensionContext.telemetry.getTelemetryReporter());
+        if (!artemisResponse) {
+            // Todo: Log in error telemetry. Runtime maintains another table for this kind of failure. We should do the same.
```
ashishchoudhary001, line 604: This is a good candidate for logging in a default/fallback cluster let's say 'US'. Here the only log would be unable to instantiate telemetry because of lack of org info.

BidishaMS: This geo not only help us to determine the org geo but also initialize 1DS SDK. I would not use 'US' in this case because post this call all data will start flowing to US cluster. There is a table Runtime uses for all such cases will leverage the same.

PR #853, opened 2024-02-20
Populate Website ID in context for desktop extension

gshivi:
Description

Populate context object in desktop extension - website ID field to be logged for every trace.
The value for this field will be logged based on the active file in vs code. 
If no active file or issue in finding root folder for the data - empty string gets logged as website id.

### DIFF
--- a/src/common/OneDSLoggerTelemetry/oneDSLogger.ts
+++ b/src/common/OneDSLoggerTelemetry/oneDSLogger.ts
@@ -11,9 +11,9 @@ import { ITelemetryLogger } from "./ITelemetryLogger";
 import { IContextInfo, IUserInfo } from "./IEventTypes";
 import { EventType, Severity } from "./telemetryConstants";
 import * as vscode from "vscode";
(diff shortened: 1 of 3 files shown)


src/common/OneDSLoggerTelemetry/oneDSLogger.ts
```diff
@@ -194,11 +194,14 @@ export class OneDSLogger implements ITelemetryLogger{
         return instrumentationSettings;
       }
 
-	/// Trace info log
-	public traceInfo(eventName:string, eventInfo?:object, measurement?: object) {
-		const event = {
-			name: OneDSCollectorEventName.VSCODE_EVENT,
-			data: {
+    /// Trace info log
+    public async traceInfo(eventName: string, eventInfo?: object, measurement?: object) {
```
BidishaMS, line 198: You can't make traceInfo as async. Neither putting this kind of condition is generic. This method should be as simple as calling 1DSLogger.

BidishaMS (changes requested):

PR #856, opened 2024-02-21
Dependabot alerts package override

tyaginidhi:
This pull request includes changes to the `package.json` file to specify the minimum required version of `npm` and to override certain package versions. 

Here are the key changes:

* [`package.json`](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L59-R60): The `engines` field now includes a specification for `npm` version `>=8.3.0`. This means that the project now requires at least version `8.3.0` of `npm` to run.
* [`package.json`](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R1125-R1128): The `overrides` field has been added to the `optionalDependencies` section. This field specifies that the versions of `jsrsasign` and `axios` should be `^11.0.0` and `^0.28.0` respectively, overriding any other specified versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1896,76 +1896,40 @@
       }
     },
     "node_modules/@fluidframework/server-services-client": {
-      "version": "0.1036.5001",
-      "resolved": "https://registry.npmjs.org/@fluidframework/server-services-client/-/server-services-client-0.1036.5001.tgz",
-      "integrity": "sha512-e+Zk2uPcds9yqlalAZ0PpfEAKPPUIpIoQb3YSm42ZVpAgQmLYIRqTpXrmTC7qdeQnSGFJUAliW4DfHBEwpSXlQ==",
+      "version": "0.1040.1000",
+      "resolved": "https://registry.npmjs.org/@fluidframework/server-services-client/-/server-services-client-0.1040.1000.tgz",
+      "integrity": "sha512-udOTa/JPCCEtRFEavEYGXMDq2taWBG/QuSnXduUvfBfIaK/ko9jmDJibN66QCWOGFhAIbpHFRRdomAZUFpnndg==",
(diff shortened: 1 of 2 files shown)


alexvy86 (commented):
I think this is correct, but since we don't use npm ourselves (we use pnpm) I'm a bit unsure if what I see in the lockfile is expected. E.g. lines 754 and 749 still have the non-overridden versions, but maybe that's normal. The way you specified overrides makes sense to me according to the docs. At least I'd double check that in your local clone of the repo, `node_modules/@fluidframework/azure-client/node_modules/@fluidframework/server-services-client/node_modules/axios/package.json` has the expected version (something that matches (^0.28.0), and the same for jsrsasign.

tyaginidhi:
> I think this is correct, but since we don't use npm ourselves (we use pnpm) I'm a bit unsure if what I see in the lockfile is expected. E.g. lines 754 and 749 still have the non-overridden versions, but maybe that's normal. The way you specified overrides makes sense to me according to the docs. At least I'd double check that in your local clone of the repo, `node_modules/@fluidframework/azure-client/node_modules/@fluidframework/server-services-client/node_modules/axios/package.json` has the expected version (something that matches (^0.28.0), and the same for jsrsasign.

Per the npm docs this is the best way to make sure any dependency which is axios and jsrsasign will only use the version we overriden with - but yeah we would rather that packages themselves use whatever version they are bringing in - this can cause trouble laterIF we miss removing these dependencies - https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides

ashishchoudhary001 (approved):

BidishaMS (approved):

PR #870, opened 2024-03-06
[Bug Fix] Initial Load Failure

ritikramuka:
This pull request includes changes to the `activate` function in the `src/web/client/extension.ts` file. The changes involve the repositioning of the instantiation of `oneDSLoggerWrapper` with `orgGeo` data fetched from `fetchArtemisData(orgId)`. The instantiation, which was previously done before the `checkMandatoryParameters` function call, is now done after the parameters check.

Here are the key changes:

* [`src/web/client/extension.ts`](diffhunk://#diff-1429bea43439e067ae9056ff3f9f71203533f7d3ca168f0d5469aa1b9a6e0360L95-L97): Moved the instantiation of `oneDSLoggerWrapper` with `orgGeo` data fetched from `fetchArtemisData(orgId)` to after the `checkMandatoryParameters` function call. This change ensures that the logger is instantiated only after the mandatory parameters have been checked. [[1]](diffhunk://#diff-1429bea43439e067ae9056ff3f9f71203533f7d3ca168f0d5469aa1b9a6e0360L95-L97) [[2]](diffhunk://#diff-1429bea43439e067ae9056ff3f9f71203533f7d3ca168f0d5469aa1b9a6e0360R111-R115)

### DIFF
--- a/src/web/client/extension.ts
+++ b/src/web/client/extension.ts
@@ -92,9 +92,6 @@ export function activate(context: vscode.ExtensionContext): void {
                         );
                     }
                 }
-                const orgId = queryParamsMap.get(queryParameters.ORG_ID) as string;
-                const orgGeo = await fetchArtemisData(orgId);
-                oneDSLoggerWrapper.instantiate(orgGeo);
                 if (
                     !checkMandatoryParameters(
                         appName,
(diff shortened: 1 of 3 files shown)


tyaginidhi (approved):

src/web/client/extension.ts
```diff
@@ -111,6 +108,11 @@ export function activate(context: vscode.ExtensionContext): void {
                     queryParamsMap,
                     context.extensionUri
                 );
+
+                const orgId = queryParamsMap.get(queryParameters.ORG_ID) as string;
```
BidishaMS, line 112: Could you please remove orgURL logging in check params API

ritikramuka: Done!

BidishaMS (approved):

BidishaMS (approved):

PR #877, opened 2024-03-15
Setting userRegion based on geo if defined, else using parsed geo

ritikramuka:
This pull request adds functionality to set the userRegion based on the geo if it is defined. If the geo is not defined, it will use the parsed geo instead.

### DIFF
--- a/src/common/OneDSLoggerTelemetry/oneDSLogger.ts
+++ b/src/common/OneDSLoggerTelemetry/oneDSLogger.ts
@@ -328,18 +328,18 @@ export class OneDSLogger implements ITelemetryLogger{
         if (envelope.data.eventName == telemetryEventNames.WEB_EXTENSION_INIT_QUERY_PARAMETERS) {
             const eventInfo = JSON.parse(envelope.data.eventInfo);
 
-            OneDSLogger.userInfo.tid = eventInfo.tenantId;
-            OneDSLogger.userRegion = eventInfo.geo ? geoMappingsToAzureRegion[eventInfo.geo.toLowerCase()].geoName : eventInfo.geo;
-            OneDSLogger.contextInfo.orgId = eventInfo.orgId;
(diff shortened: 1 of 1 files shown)


BidishaMS (approved):

src/common/OneDSLoggerTelemetry/oneDSLogger.ts
```diff
@@ -324,22 +325,24 @@ export class OneDSLogger implements ITelemetryLogger{
...
-            OneDSLogger.contextInfo.correlationId = JSON.parse(envelope.data.eventInfo).referrerSessionId;
-            OneDSLogger.contextInfo.referrer = JSON.parse(envelope.data.eventInfo).referrer;
-            OneDSLogger.contextInfo.envId = JSON.parse(envelope.data.eventInfo).envId;
-            OneDSLogger.contextInfo.referrerSource = JSON.parse(envelope.data.eventInfo).referrerSource;
-            OneDSLogger.contextInfo.orgGeo = JSON.parse(envelope.data.eventInfo).orgGeo;
+        if (envelope.data.eventName == telemetryEventNames.WEB_EXTENSION_INIT_QUERY_PARAMETERS) {
+            const eventInfo = JSON.parse(envelope.data.eventInfo);
+
+            OneDSLogger.userInfo.tid = eventInfo.tenantId;
+            OneDSLogger.userRegion = eventInfo.geo ? geoMappingsToAzureRegion[eventInfo.geo.toLowerCase()].geoName : eventInfo.geo;
```
ashishchoudhary001, line 332: What are we doing here?

If eventInfo.geo doesn't exist, we still end up logging that? In such a scenario what would be logged here?

BidishaMS: Quite unlikely eventInfo.geo will have undefined values because in Studio it is always defined. In any scenario it is undefined we should log empty

image

ashishchoudhary001 (approved):

PR #908, opened 2024-04-15
powerplatform-vscode Localization hand-back [CTAS - RunID=20240415-221949-bg61sfv201]

toshio-msft:
Translation handback check-in (PR Created by automation).


tyaginidhi:
these changes went in as part of another recent PR - no chnages here to commit

PR #910, opened 2024-04-18
[Bug Fix] Co-Presence for non-localized content

ritikramuka:
Bug Fixes in this PR: 
1. Co-Presence was not triggering for non-localized content (i.e. entities other than webpage), this PR addresses the issue and fixes the same.
image

2. If no other user is present on the page: People on this site should be 0
image

### DIFF
--- a/src/web/client/utilities/fileAndEntityUtil.ts
+++ b/src/web/client/utilities/fileAndEntityUtil.ts
@@ -31,13 +31,16 @@ export function getFileEntityName(fileFsPath: string) {
 
 export function getFileRootWebPageId(fileFsPath: string) {
     const entityId = getFileEntityId(fileFsPath);
-    return (
+
+    const rootWebPageId =
         (WebExtensionContext.entityDataMap.getEntityMap.get(entityId)
             ?.rootWebPageId as string) ??
         (WebExtensionContext.getVscodeWorkspaceState(fileFsPath)
             ?.rootWebPageId as string) ??
-        ""
-    );
+        "";
+
+    // If rootWebPageId is not present, return entityId as content is not localized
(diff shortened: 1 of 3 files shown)


src/web/client/utilities/fileAndEntityUtil.ts
```diff
@@ -31,13 +31,16 @@ export function getFileEntityName(fileFsPath: string) {
...
+    const rootWebPageId =
         (WebExtensionContext.entityDataMap.getEntityMap.get(entityId)
             ?.rootWebPageId as string) ??
         (WebExtensionContext.getVscodeWorkspaceState(fileFsPath)
             ?.rootWebPageId as string) ??
-        ""
-    );
+        "";
+
+    // If rootWebPageId is not present, return entityId as content is not localized
```
tyaginidhi, line 42: what do you mean by this comment? Per this code - in case of webpage you have a rootPageId which will not be the case for other entities

tyaginidhi (approved):

src/web/client/utilities/fileAndEntityUtil.ts
ritikramuka: Webpages are localised, hence we have two different guid content page id root page id, so as other entities are not localised therefore storing same guid in root page id and entity page id.

amitjoshi438 (approved):

src/web/client/utilities/fileAndEntityUtil.ts
tyaginidhi: Be careful here - in case of content snippets as well content can be localized

ritikramuka: yes that's why it is like a fallback scenario, that if we don't receive any rootPageId then make it same as entity Id

PR #911, opened 2024-04-19
[Telemetry] Adding x-ms-user-agent in Dataverse calls to check the client  surface (Vscode Web/Desktop)

BidishaMS:
[Telemetry] Adding x-ms-user-agent in Dataverse calls to check the client  surface (Vscode Web/Desktop)


amitjoshi438 (approved):

amitjoshi438 (approved):

src/common/copilot/dataverseMetadata.ts
```diff
@@ -25,6 +25,7 @@ export async function getEntityColumns(entityName: string, orgUrl: string, apiTo
             headers: {
                 'Content-Type': "application/json",
                 Authorization: `Bearer ${apiToken}`,
+                "x-ms-user-agent": "Vscode-Desktop"
```
joem-msft, line 28: This is a great idea.
The expected format of this string is the same as the [User-Agent header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent). Namely:
`x-ms-user-agent: <product> / <product-version> <comment>`

1. Let's add the version of the product too. It must be the version of the 'product'
1. For 'product', Should it actually be the name or id of the extension? If not, and we want it to be the VSCode product, can we set the product value to something we get from the VSCode SDK instead?
2. In the 'comment' part, is where extra information is usually set. e.g. the desktop vs web would be more appropriate there. e.g. `VSCode/1.88.1 (Desktop; other; another)`

src/common/copilot/dataverseMetadata.ts
```diff
@@ -57,6 +58,7 @@ export async function getFormXml(entityName: string, formName: string,  orgUrl:
...
```
joem-msft, line 61: Looks like this header constant is being used in multiple places. Consider moving it's value to a reusable function, with different implementation per Desktop/Web.
e.g. `composeXMsUserAgent`

joem-msft (changes requested):
only needing the product version to be added to the x-ms-user-agent header values.
all other comments are recommendations.

src/common/copilot/dataverseMetadata.ts
tyaginidhi: +1 to this comment  @BidishaMS

tyaginidhi (changes requested):

src/common/copilot/dataverseMetadata.ts
BidishaMS: Done

src/common/copilot/dataverseMetadata.ts
BidishaMS: Done

src/common/Utils.ts
```diff
@@ -5,7 +5,7 @@
 
 
 import * as vscode from "vscode";
-import { EXTENSION_ID, SETTINGS_EXPERIMENTAL_STORE_NAME } from "../client/constants";
+import { EXTENSION_ID,EXTENSION_NAME, SETTINGS_EXPERIMENTAL_STORE_NAME } from "../client/constants";
```
tyaginidhi, line 8: nit: spacing not accurate

tyaginidhi (approved):

src/common/Utils.ts
BidishaMS: fixed

src/common/Utils.ts
```diff
@@ -123,3 +123,12 @@ export function isCustomTelemetryEnabled():boolean {
         .get(CUSTOM_TELEMETRY_FOR_POWER_PAGES_SETTING_NAME);
     return isCustomTelemetryEnabled as boolean;
 }
+
+export function getUserAgent(): string {
+    const userAgent = "{product} / {product-version} {comment}";
```
joem-msft, line 128: there should not be any space between the product, '/' char and version. The space after the version is significant.
"{product}/{product-version} {comment}"

src/common/Utils.ts
```diff
@@ -123,3 +123,12 @@ export function isCustomTelemetryEnabled():boolean {
...
+
+    return userAgent
+        .replace("{product}", EXTENSION_NAME)
+        .replace("{product-version}", getExtensionVersion())
+        .replace("{comment}", getExtensionType()+'; '+getExtensionVersion());
```
joem-msft, line 133: See how browsers do this. The comment, if specified is usually in parentheses. This makes it so spaces can be inside the comment.
i.e. `"myproduct/0.1.0-dev (extVer:5.0; otherThing:value; flagValue)"`

joem-msft, line 133: Looks like the extension version is there twice. Remove from the comment?

joem-msft (changes requested):

src/common/Utils.ts
BidishaMS: removed

BidishaMS: put it in parenthesis

joem-msft (approved):

PR #1026, opened 2024-08-30
BugItem-4312228: changed the reorder of powerpagesNavigation

ashwani123p:
Bug: https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/4312228

In the web version of VS Code, the authentication prompt currently appears twice during launch. The first prompt is triggered by the powerPagesNavigation function, which internally calls powerPlatformAPIAuthentication with a scope limited to api.powerplatform. The second prompt occurs when readDirectory is invoked, as it uses dataverseAuthentication internally with a broader default scope.

The issue arises because when the user is first authenticated through powerPlatformAPIAuthentication, the session created is not recognized by dataverseAuthentication due to the scope limitations. However, if the user is authenticated with dataverseAuthentication first, the resulting session is recognized by powerPlatformAPIAuthentication as well. So changing the order of powerPagesNavigation function

### DIFF
--- a/src/web/client/extension.ts
+++ b/src/web/client/extension.ts
@@ -139,8 +139,6 @@ export function activate(context: vscode.ExtensionContext): void {
 
                                 processWalkthroughFirstRunExperience(context);
 
-                                powerPagesNavigation();
-
                                 await vscode.window.withProgress(
                                     {
                                         location: vscode.ProgressLocation.Notification,
@@ -165,6 +163,8 @@ export function activate(context: vscode.ExtensionContext): void {
                                         processWillStartCollaboration(context);
                                     }
                                 );
+
+                                powerPagesNavigation();
 
                                 await NPSService.setEligibility();
                                 if (WebExtensionContext.npsEligibility) {


src/web/client/extension.ts
```diff
@@ -166,6 +164,8 @@ export function activate(context: vscode.ExtensionContext): void {
                                     }
                                 );
 
+                                powerPagesNavigation();
```
tyaginidhi, line 167: what is the change in experience with this? the Actions hub shows "no data provider regsitered" till all files are fetched?
 
Alternately, can we await this function instead and then proceed to fetching files? What is the experience in that case? Is the delay too long?

ashwani123p: image
Until the user is authenticated and before all files load, this message will be shown if the user has opened the Power Pages actions.
Resolving this comment, as this is not hard requirement, will come up later for better experience

tyaginidhi: this option looks good - we should consider long term experience with design on this

amitjoshi438 (approved):

PR #1031, opened 2024-09-05
Add Power_Platform_Tools_Feed to supported Nuget feeds.

sanxsth:
The PR contains the following changes:
1. Adds Power_Platform_Tools_Feed feed to the supported feeds.
2. Introduces a new pipeline argument 'useFeed' that can be used to choose to Nuget Feed during build.

### DIFF
--- a/gulpfile.mjs
+++ b/gulpfile.mjs
@@ -42,6 +42,7 @@ const packagedir = path.resolve('./package');
 const feedPAT = argv.feedPAT || process.env['AZ_DevOps_Read_PAT'];
 const isOfficialBuild = argv.isOfficialBuild && argv.isOfficialBuild.toLowerCase() == "true";
 const isPreviewBuild = argv.isPreviewBuild && argv.isPreviewBuild.toLowerCase() == "true";
+const feedName = argv.useFeed || 'nuget.org';
 
 async function clean() {
     (await pslist())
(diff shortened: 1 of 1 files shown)


gulpfile.mjs
```diff
@@ -42,6 +42,7 @@ const packagedir = path.resolve('./package');
 const feedPAT = argv.feedPAT || process.env['AZ_DevOps_Read_PAT'];
 const isOfficialBuild = argv.isOfficialBuild && argv.isOfficialBuild.toLowerCase() == "true";
 const isPreviewBuild = argv.isPreviewBuild && argv.isPreviewBuild.toLowerCase() == "true";
+const feedName = argv.useFeed || 'nuget.org';
```
tyaginidhi, line 45: can we check once if nuget is the right default in case useFeed param is missing? or do we want to take the 2 other alternate present in gulpfile based on the right params

tyaginidhi (approved):

gulpfile.mjs
sanxsth: Nuget.org was the previous default in place. The useFeed now acts like an override.

sanxsth: image
We had explicitly set 'feedName' to 'nuget.org' as default.

amitjoshi438 (approved):

PR #1091, opened 2025-01-21
Web Extension: Remove dependency on website preview URL

priyanshu92:
This pull request introduces several changes to improve the validation and handling of website preview URLs in the codebase. The most significant changes include modifying the `validateWebsitePreviewURL` function to return an object with the URL and its validity status, updating the `PowerPagesNavigationProvider` class to use this new validation result, and removing the `WEBSITE_PREVIEW_URL` query parameter.

### Changes to URL validation:

* [`src/web/client/utilities/commonUtil.ts`](diffhunk://#diff-0571fa597154eb22907e2fbbd10adc9ccc04cba979a11f3c571f94a54abcf179L313-R313): Modified the `validateWebsitePreviewURL` function to return an object with `websiteUrl` and `isValid` properties instead of a boolean.

### Updates to `PowerPagesNavigationProvider` class:

* [`src/web/client/webViews/powerPagesNavigationProvider.ts`](diffhunk://#diff-96d103f6c47a0b6a7fee791b25952d72eefa3b5e6af58bf38c8d6bb25663683eL9-R17): Updated the `isWebsitePreviewURLValid` property and the `previewPowerPageSite` method to use the new validation result object. This ensures that the correct URL and its validity are checked before proceeding.

### Removal of unused query parameter:

* [`src/web/client/common/constants.ts`](diffhunk://#diff-8769a22aef8333b09b6f98fe24f9c95ece22948a556d3d9f64528f4a6642add3L99): Removed the `WEBSITE_PREVIEW_URL` query parameter from the `queryParameters` enum as it is no longer needed.


loc/translations-export/vscode-powerplatform.xlf
```diff
@@ -548,8 +545,8 @@ The {3} represents Dataverse Environment's Organization ID (GUID)</note>
       <source xml:lang="en">Auth Profiles</source>
     </trans-unit>
     <trans-unit id="microsoft-powerapps-portals.walkthrough.saveConflict.description">
-      <source xml:lang="en">Avoid accidental overwrites when you try to save outdated code in VS Code for the Web. You can compare the changes side-by-side and decide to accept or revert the changes.
-
+      <source xml:lang="en">Avoid accidental overwrites when you try to save outdated code in VS Code for the Web. You can compare the changes side-by-side and decide to accept or revert the changes. 
+ 
 To learn more, visit [Prevent accidental overwrites](command:powerplatform-walkthrough.saveConflict-learn-more).</source>
```
ashishchoudhary001, line 550: Do we have any change in this file? If not, could we revert it please?

priyanshu92: This is automatically showing up in diff when running `gulp ci`.

src/web/client/utilities/commonUtil.ts
```diff
@@ -310,7 +310,7 @@ export function getRangeForMultilineMatch(text: string, pattern: string, index:
     return range;
 }
 
-export async function validateWebsitePreviewURL(): Promise<boolean> {
+export async function validateWebsitePreviewURL(): Promise<{ websiteUrl: string, isValid: boolean }> {
```
sanxsth: NIT: maybe we can have a name like 'getValidWebsitePreviewURL' or similar to show that we are not just validating the URL, rather returning the URL value as well.

sanxsth (approved):

src/web/client/utilities/commonUtil.ts
priyanshu92: Done

PR #1140, opened 2025-02-27
[PowerPages] [Actions Hub] Add support for other sites in Actions Hub

amitjoshi438:
Introduce functionality to handle additional sites in the Actions Hub, along with updates to related interfaces and tests to ensure proper handling of these new sites.

image

### DIFF
--- a/src/client/power-pages/actions-hub/ActionsHubCommandHandlers.ts
+++ b/src/client/power-pages/actions-hub/ActionsHubCommandHandlers.ts
@@ -271,7 +271,7 @@ export const uploadSite = async (siteTreeItem: SiteTreeItem) => {
  * @param knownSiteIds Map of site IDs that should be excluded from results
  * @returns Array of site information objects for sites found in the parent folder
  */
-export function findOtherSites(knownSiteIds: Map<string, boolean>): IOtherSiteInfo[] {
+export function findOtherSites(knownSiteIds: Map<string, boolean>, fsModule = fs, yamlModule = yaml): IOtherSiteInfo[] {
     // Get the workspace folders
     const workspaceFolders = vscode.workspace.workspaceFolders;
     if (!workspaceFolders || workspaceFolders.length === 0) {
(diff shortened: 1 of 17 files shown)


src/client/power-pages/actions-hub/ActionsHubCommandHandlers.ts
```diff
@@ -259,3 +265,98 @@ export const uploadSite = async (siteTreeItem: SiteTreeItem) => {
...
+        for (const dir of directories) {
+            const websiteYamlPath = path.join(dir, WEBSITE_YML);
+
+            if (fsModule.existsSync(websiteYamlPath)) {
+                try {
+                    // Parse website.yml to get site details
+                    const yamlContent = fsModule.readFileSync(websiteYamlPath, UTF8_ENCODING);
+                    const websiteData = yamlModule.load(yamlContent) as WebsiteYaml;
+
+                    const websiteId = websiteData?.adx_websiteid;
```
priyanshu92, line 308: I think we should be able to replace this with `getWebsiteRecordId` from `WorkspaceInfoFinderUtil.ts`

src/client/power-pages/actions-hub/ActionsHubCommandHandlers.ts
```diff
@@ -259,3 +265,98 @@ export const uploadSite = async (siteTreeItem: SiteTreeItem) => {
...
+        );
+        return [];
+    }
+}
+
+export function createKnownSiteIdsMap(
+    activeSites: IWebsiteDetails[] | undefined,
+    inactiveSites: IWebsiteDetails[] | undefined
+): Map<string, boolean> {
+    const knownSiteIds = new Map<string, boolean>();
```
priyanshu92, line 345: Do we need a `Map` here? Since we are just checking the existense I guess we can use `Set` instead. What do you think?

priyanshu92 (approved):
2 small things otherwise LGTM

src/client/power-pages/actions-hub/ActionsHubCommandHandlers.ts
amitjoshi438: Right, it will be more memory efficient.

PR #1180, opened 2025-04-14
[PowerPages][Web] PPHome Launch Error Page fix for V1

amitjoshi438:
Introduce a REFERRER enum and update PortalsFS to manage referrer logic based on the portal version, improving handling of default file loading scenarios.

### DIFF
--- a/src/web/client/common/constants.ts
+++ b/src/web/client/common/constants.ts
@@ -127,6 +127,10 @@ export enum GraphService {
     GRAPH_PROFILE_PICTURE = "profilePicture",
 }
 
+export enum REFERRER{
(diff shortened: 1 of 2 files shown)


src/web/client/common/constants.ts
```diff
@@ -127,6 +127,10 @@ export enum GraphService {
     GRAPH_PROFILE_PICTURE = "profilePicture",
 }
 
+export enum REFERRER{
```
priyanshu92, line 130: nit: space before `{`

priyanshu92 (approved):

sanxsth (approved):

PR #1253, opened 2025-08-12
[PowerPages] Add Telemetry fo BL API autocomplete support

amitjoshi438:
Implement Server API autocomplete functionality, including namespaces and properties, with comprehensive documentation and tests. Introduce telemetry support for tracking autocomplete events and add feature gating for conditional activation. Remove outdated test suites and enhance integration testing workflows.


src/web/client/extension.ts
```diff
@@ -154,6 +155,15 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
                                         const { enableBLChanges } = EnableBLChanges.getConfig() as { enableBLChanges?: boolean };
                                         if (!serverApiAutocompleteInitialized && enableBLChanges) {
+                                            // Set telemetry context for Server API autocomplete events
+                                            setServerApiTelemetryContext({
+                                                tenantId: queryParamsMap.get(queryParameters.TENANT_ID) as string,
+                                                envId: WebExtensionContext.environmentId,
+                                                userId: WebExtensionContext.userId,
+                                                orgId: orgId,
+                                                geo: WebExtensionContext.geoName,
+                                                extType: 'web'
```
priyanshu92: Do we need `extType`? I think this is already part of telemetry.

src/common/intellisense/serverApiTelemetryEventNames.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ */
+
+/**
+ * Telemetry event names for Server API autocomplete functionality
+ * These events track the usage and performance of Server API IntelliSense features
+ * in Power Pages JavaScript development within VS Code
+ */
+export enum ServerApiTelemetryEventNames {
```
priyanshu92: File name casing can be updated to match others

priyanshu92 (approved):
Is there an event to know when user accepts something suggested by intellisense? If so, we should log that as well.

vipulbandi (approved):

src/web/client/extension.ts
amitjoshi438: Oh okay, because it's shared code I thought of adding it..but if it's part of telemetry infra we can remove it.

PR #1352, opened 2025-11-19
Enhance authentication session handling

priyanshu92:
- 🔧 Updated session retrieval to include offline access scope.
- 🔄 Refactored session handling to create sessions if none exist.
- 🛠️ Improved error handling for access token retrieval.

-Priyanshu

### DIFF
--- a/src/common/ecs-features/ecsFeatureGates.ts
+++ b/src/common/ecs-features/ecsFeatureGates.ts
@@ -34,7 +34,7 @@ export const {
     description: 'Enable ProDev Copilot',
     fallback: {
         "capiSupportedProDevCopilotGeoList": "us,au,uk,eu,in",
-        "unsupportedProDevCopilotGeoList": "ca",
+        "unsupportedProDevCopilotGeoList": "",
         "capiSupportedProDevCopilotGeoWithCrossGeoDataFlow": "eu,se,ch,fr,de,no"
     },
 });


src/common/services/AuthenticationProvider.ts
```diff
@@ -89,10 +89,10 @@ export async function intelligenceAPIAuthentication(sessionID: string, orgId: st
     let user = '';
     let userId = '';
     try {
-        const session = await vscode.authentication.getSession(PROVIDER_ID, [`${INTELLIGENCE_SCOPE_DEFAULT}`], { silent: true });
+        let session = await vscode.authentication.getSession(PROVIDER_ID, [`${INTELLIGENCE_SCOPE_DEFAULT}`], { silent: true });
 
         if (!session) {
-            return { accessToken, user, userId };
+            session = await vscode.authentication.getSession(PROVIDER_ID, [`${INTELLIGENCE_SCOPE_DEFAULT}`], { createIfNone: true });
```
amitjoshi438, line 95: Will it cause multiple auth prompts?

amitjoshi438 (approved):

src/common/services/AuthenticationProvider.ts
priyanshu92: Ideally it shouldn't but if somehow the token acquisition fails in silent mode then it will prompt the user. This is still better than not showing the content due to token failure.

PR #1363, opened 2025-12-09
[MDiff] Add compare with local action in Power Pages Actions

priyanshu92:
This PR adds a new action to Power Pages Actions called **Compare With Local**. This feature allows a user to compare the Power Pages metadata for a site which is opened in VS Code Desktop workspace with that of any site in the connected environment in Power Pages Actions.

Comparing the metadata updates the Power Pages Actions menu with a tree view of the files which are different in both sites allowing user to open the diff view to see the difference.

### DIFF
--- a/src/common/ecs-features/ecsFeatureGates.ts
+++ b/src/common/ecs-features/ecsFeatureGates.ts
@@ -122,11 +122,21 @@ export const {
 });
 
 export const {
+    feature: EnableBlogSupport
+} = getFeatureConfigs({
+    teamName: PowerPagesClientName,
+    description: 'Enable blog file support in VSCode (web & desktop)',
+    fallback: {
+        enableBlogSupport: false,
+    }
+});
+
+export const {
     feature: EnableMetadataDiff
 } = getFeatureConfigs({
     teamName: PowerPagesClientName,
-    description: 'Enable Metadata Diff comparison in VS Code',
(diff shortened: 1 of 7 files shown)


src/client/power-pages/actions-hub/handlers/metadata-diff/CompareWithLocalHandler.ts
```diff
@@ -0,0 +1,193 @@
...
+                    siteId: siteTreeItem.siteInfo.websiteId
+                });
+                vscode.window.showInformationMessage(Constants.Strings.NO_DIFFERENCES_FOUND);
+                MetadataDiffContext.clear();
+            } else {
+                traceInfo(Constants.EventNames.ACTIONS_HUB_COMPARE_WITH_LOCAL_COMPLETED, {
+                    methodName: compareWithLocal.name,
+                    siteId: siteTreeItem.siteInfo.websiteId,
+                    totalDifferences: comparisonResults.length.toString(),
+                    modifiedFiles: comparisonResults.filter(r => r.status === "modified").length.toString(),
```
amitjoshi438, line 181: nit: maybe we can move these to constants as they are used at multiple places

priyanshu92: Have done it in subsequent PRs

PR #1409, opened 2025-12-31
[MDiff] Update metadata diff labels and add clear functionality

priyanshu92:
✨ Renamed "Metadata Diff" to "Site Comparison" for clarity. ✨ Added confirmation dialogs for clearing metadata diff results. ✨ Implemented functionality to clear individual site comparisons. ✨ Updated localization files with new labels and messages. ✨ Added tests for new clear functionality and site comparison behavior.


src/client/test/Integration/power-pages/actions-hub/tree-items/metadata-diff/MetadataDiffGroupTreeItem.test.ts
```diff
@@ -56,7 +56,7 @@ describe("MetadataDiffGroupTreeItem", () => {
             MetadataDiffContext.setResults(results, "Test Site", "Local Test Site", "Test Environment", "test-website-id", "test-environment-id");
             const treeItem = new MetadataDiffGroupTreeItem();
 
-            expect(treeItem.label).to.equal("Metadata Diff");
+            expect(treeItem.label).to.equal("Site Comparison");
```
amitjoshi438, line 59: nit: Maybe we can use constant here also as it's already defined for this string.

amitjoshi438 (approved):

src/client/test/Integration/power-pages/actions-hub/tree-items/metadata-diff/MetadataDiffGroupTreeItem.test.ts
priyanshu92: It's better to use actual strings in test to ensure that the value in constant file is correct. If there is some change in constant file then this test will allow us to either update the test or detect unintended change.

PR #1561, opened 2026-04-23
Fix CLA blocking in Dependabot auto-fix workflow

priyanshu92:
## Context

The first scheduled run of the Dependabot auto-fix workflow (added in #1558) opened PR #1559 but got stuck on the Microsoft CLA bot. Root cause: Microsoft's CLA bot keys off **PR author**, not commit author, and refuses to evaluate when the PR author is a GitHub App. The App-owned PR stayed `license/cla: queued` even after posting `@microsoft-github-policy-service agree` and re-authoring the commit; the only way to get CLA to pass was closing PR #1559 and reopening the same branch as PR #1560 under a human author.

## Fix

Use the existing `COPILOT_CLI_PAT` for **all** GitHub write operations in the workflow. Since the PAT is owned by a Copilot-licensed, CLA-signed human, the PR author becomes that user and CLA passes automatically on every weekly run.

### Changes
- Remove the `actions/create-github-app-token@v1` step and the `POWER_PAGES_PUBLIC_GITHUB_APP_PRIVATE_KEY` reference from this workflow (the App token is not used any more — the other workflows that rely on it are unaffected).
- Add a `Resolve PAT owner identity` step that calls `gh api user` and configures `git user.name` / `user.email` to the PAT owner, so commit authors are CLA-clean too.
- Point `GH_TOKEN` at `COPILOT_CLI_PAT` for checkout, `git push`, and `gh pr create`.
- Tell the agent to **not** add any `Co-authored-by:` trailers (extra identities also trip the CLA bot).

### Required PAT permissions (updated)
`COPILOT_CLI_PAT` must now also include **Contents: Read/Write** and **Pull requests: Read/Write** in addition to the original Copilot Requests + Dependabot alerts reads. Please update the existing secret before the next scheduled run.

## Test plan
After merge, trigger the workflow manually via `workflow_dispatch`; verify the resulting PR is authored by the PAT owner and the `license/cla` check turns green without manual intervention.

### DIFF
--- a/.github/workflows/dependabot-autofix.yml
+++ b/.github/workflows/dependabot-autofix.yml
@@ -10,21 +10,28 @@
 #
 # Required secrets
 # ------------------------------------------------------------------------
-#   POWER_PAGES_PUBLIC_GITHUB_APP_PRIVATE_KEY
-#     Private key of the existing GitHub App (app-id 2740120) that is
-#     already used by loc-update.yml / translations-export.yml. The App
-#     installation on this repo must have:
-#       - Contents: Read/Write
-#       - Pull requests: Read/Write
-#     Commits and the PR are authored as github-actions[bot] via this App.
-#
 #   COPILOT_CLI_PAT
-#     Fine-grained PAT from a user with a Copilot license. Needed because
-#     GitHub App tokens cannot authenticate Copilot CLI (Copilot licenses
-#     are per-user). Required permissions:
-#       - Account: Copilot Requests: Read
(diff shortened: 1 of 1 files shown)


priyanshu92:
Closing — switching PR author to a human would prevent the repo admin from self-approving the auto-fix PR. Keeping the App as PR author and pursuing a CLA allow-list for `power-pages-github-app[bot]` instead (see comment on the PR). Branch will be deleted.

PR #231, opened 2022-07-07
PCF Debugger: Add initial debugger code

felixSchober:
Adds a debugger option for PCF controls.
This works by attaching the edge debugger to a puppeteer session which opens a model driven app, intercepts network requests to the bundle and responds with the local version of the bundle.

To use the debugger with any PCF control you need to perform the following steps:

1. **Enable Feature**: Make sure "_Power Platform -> Experimental: Enable Pcf Debugging Features_" is enabled. After enabling the feature, you need to restart vscode.
image

2. **tsconfig sourcemaps**: Open `tsconfig.json` and add `"sourceMap": true` under `compilerOptions`. Your file should look something like this:

  ```json
  {
      "extends": "./node_modules/pcf-scripts/tsconfig_base.json",
      "compilerOptions": {
          "sourceMap": true,
      }
  }
  ```

3. **Add custom webpack config**: To enable the custom webpack config feature flag you need to create a file called `featureconfig.json` in your project root. Use the following contents:

```json
{
    { "pcfAllowCustomWebpack": "on" }
}
```

4. **Generate source maps**: To generate source maps with webpack, create a new file called `webpack.config.js` in your project root. Use the following contents:

```js
const customConfig = {
  // watch: true, // uncomment this line to enable webpack watch mode
  devtool: "inline-source-map"
}
module.exports = customConfig;
```

5. **Add a launch config**: Create a new `launch.json` file under `.vscode`. If the folder `.vscode` does not yet exist, you can create it.
Click on "_Add Configuration..._" and select for "_Debug PCF Control_" from the dropdown. (If you can't find the entry the you might have to restart vscode.)

image

A configuration for the [Sample: TableControl](https://github.com/microsoft/PowerApps-Samples/tree/master/component-framework/TableControl) is listed below:
```json
{
  "version": "0.2.0",
  "configurations": [
    {
      "type": "powerplatform-vscode.debug",
      "request": "launch",
      "name": "Launch Control",
      "url": "https://YOUR_ORG_URL.crm.dynamics.com/main.aspx",
      "webRoot": "${workspaceFolder}",
      "renderFullScreen": false,
      "controlName": "cc_SampleNamespace.TableControl",
      "file": "out/controls/TableControl/bundle.js",
      "tabName": "Sample Control Tab"
    }
  ]
}
```

[...]

PR #1638, opened 2026-07-29
Fix Dependabot alerts in tar and brace-expansion

priyanshu92:
Closes two open Dependabot alerts, plus the macOS CI break that surfaced on this branch.

### Security fixes

- **`tar` 7.5.19 -> 7.5.21** (alert #230, medium). Uncontrolled recursion in `mapHas`/`filesFilter` lets a crafted long-path tar with member selection blow the stack in a way the caller cannot catch. Picked up by `npm update`; the existing `^7.5.16` override already permitted the new version.
- **`brace-expansion` 5.0.7 -> 5.0.8** (alert #231, high). Unbounded expansion length can exhaust memory and crash the process. This one is pinned to an exact version in the `overrides` block, so `package.json` had to move in lockstep with the lock file.

### macOS CI fix

The macOS leg of the build was already broken on `main` before this branch, it just showed up here first. VS Code 1.131.0 (released Jul 28) renamed the macOS app bundle executable from `Contents/MacOS/Electron` to `Contents/MacOS/Code`, and `@vscode/test-electron` 2.5.2 hardcodes the old name, so every macOS run died immediately after download with:

```
spawn .../Visual Studio Code.app/Contents/MacOS/Electron ENOENT
```

Linux and Windows were unaffected because their launcher paths did not change.

`@vscode/test-electron` 3.1.0 resolves the executable from the bundle contents instead of assuming a fixed name, so this bumps to it. That release declares `engines.node >= 22`, so the CI jobs that install and run it move from Node 20 (already past end of life) to Node 22. I also refreshed the `'stable'` pin comments in the two `runTest.ts` files that carried a workaround note for the older resolver.

### Notes for reviewers

The internal npm proxy has not mirrored `brace-expansion@5.0.8` yet (the tarball 404s), so I could not install it locally to let npm write the lock entry. The `resolved` and `integrity` values were taken from the upstream registry metadata and cross-checked against 1512 public `package-lock.json` files that already pin 5.0.8, so `npm ci` against registry.npmjs.org resolves normally. Local installs on machines behind the proxy will fail on that package until the mirror catches up.

Separately, `npm update` and `npm install` rewrote the touched lock entries to `ms-feed-*.pkgs.visualstudio.com` URLs with sha1 integrity. Those were normalized back to registry.npmjs.org with sha512 before committing, which is why the diff stays limited to the real version bumps.

### Validation

Locally on macOS: reproduced the `ENOENT` failure, confirmed VS Code launches again after the `@vscode/test-electron` bump, and `npm run dist` (compile, vsix package, lint, 106 unit tests) passes.

### DIFF
--- a/.github/workflows/PullRequest.yml
+++ b/.github/workflows/PullRequest.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Setup Node.js environment
         uses: actions/setup-node@v6
         with:
-          node-version: 20
+          node-version: 22
 
       # - name: Install dependencies
       #   run: npm i -g npm
@@ -104,7 +104,7 @@ jobs:
       - name: Setup Node.js environment
         uses: actions/setup-node@v6
         with:
-          node-version: 20
+          node-version: 22
 
       - name: Install dependencies
         run: npm ci
--- a/package-lock.json
+++ b/package-lock.json
@@ -4742,9 +4742,9 @@
       "dev": true
     },
     "node_modules/brace-expansion": {
-      "version": "5.0.7",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-5.0.7.tgz",
-      "integrity": "sha512-7oFy703dxfY3/NLxC1fh2SUCQ0H9rmAY+5EpDVfXjUTTs+HEwR2nYaqLv+GWcTsumwxPfiz6CzCNkwXwBUwqCA==",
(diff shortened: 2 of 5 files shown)


priyanshu92:
Split into two independent PRs so the macOS CI fix isn't blocked behind the security review: #1639 (macOS integration tests broken by VS Code 1.131.0) and #1640 (Dependabot alerts in tar and brace-expansion).

PR #1648, opened 2026-08-11
URI handler: agent-host detection utility (dark)

amitjoshi438:
## Summary
- Adds a pure Node utility that probes `copilot --version` and `claude --version` on `PATH` and returns typed installed/version results.
- Reuses `URI_CONSTANTS.AGENT_HOST_VALUES` as the single source of truth and keeps the adjustable probe commands in a private local map.
- Accepts an injected async probe runner for fast unit tests without stubbing `child_process`; production uses a private promisified `exec` adapter.
- Probes both supported hosts in parallel with `Promise.all` and treats any probe failure as a normal not-installed result.

## Dark behavior
- Emits no telemetry.
- Is not imported or wired into any URI handler.
- Does not change any ECS gate or user-facing behavior.

## Probe assumption
The exact `--version` subcommands are an intentionally centralized, adjustable assumption for this foundation-only slice.

## Validation
- `node node_modules/gulp/bin/gulp.js lint`
- `npm run compile-tests`
- `npm test` (110 passing)
- `npm run build`


priyanshu92 (approved):

src/client/uriHandler/utils/detectAgentHost.ts
```diff
@@ -0,0 +1,85 @@
...
+    [AgentHost.Copilot]: 'copilot --version',
+    [AgentHost.Claude]: 'claude --version'
+};
+
+const AGENT_HOSTS: AgentHost[] = Object.values(AgentHost);
+const AGENT_HOST_PROBE_TIMEOUT_MS = 10000;
+const execAsync = promisify(exec);
+
+const defaultRunProbe: AgentHostProbe = async (command) => {
+    const { stdout } = await execAsync(command, { timeout: AGENT_HOST_PROBE_TIMEOUT_MS });
```
priyanshu92: Separate these into commands and arguments. Instead of running `claude --version` run `claude` with `--version` as argument

priyanshu92 (changes requested):

src/client/uriHandler/utils/detectAgentHost.ts
priyanshu92: @copilot How to fix this security issue?

amitjoshi438: Addressed in 7bd0ba62. The probe definitions now separate the executable from its arguments, and the default runner uses `execFile(command, args)` rather than `exec`. This invokes `copilot`/`claude` directly with `['--version']`, avoiding shell parsing while preserving the timeout and injected-runner testability.

priyanshu92 (approved):

